PR #7, opened 2026-07-15
raise claude agent sdk stream buffer

Sonorous281:
The SDK caps each stream-json line at 1 MB, a turn with several tool results on one line can cross it and crash the reader mid-episode. 8 MB gives headroom.


qurakchin:
Is this used to send a big and clear img to the claude code?

PR #8, opened 2026-07-15
make vla server action_chunk and num_steps configurable

Sonorous281:
Were hardcoded to 5/5; LIBERO-10 (long) runs at 10/4. Add --action-chunk and --num-steps, defaults 5/5 so existing launches are unchanged.

### DIFF
--- a/robots/libero/vla_server.py
+++ b/robots/libero/vla_server.py
@@ -66,7 +66,11 @@ class PredictRequest(BaseModel):
 CHECKPOINT_PATH = get_pi05_checkpoint_path()
 
 
-def build_model_cfg(model_path: str | None = None) -> Any:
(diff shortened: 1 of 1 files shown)


qurakchin (commented):
I think it's better to add args in server function and add the config in client.

Sonorous281:
> I think it's better to add args in server function and add the config in client.

It's just an optional flag. Doing it per-request from client would also need changes in the rlinf model code, since that's where action_chunk  and num_steps actually get read from self.config, and that path is shared by all openpi evals. A run stays on one value anyway, so it doesn't really buy much.

qurakchin:
It's not used now, and the flag is not good design. i trans it to a draft pr.

PR #20, opened 2026-07-20
fix: correct Claude turn accounting

wilburx813:
## Summary

- Correctly count unique top-level Claude turns.
- Avoid SDK `num_turns` overriding the observed count.
- Keep consuming the stream after `finish` to capture final usage and cost.


qurakchin (commented):
i think max_turns should be send to ClaudeAgentOptions

PR #35, opened 2026-07-22
Clean up libero prompt organization

wilburx813:
## Summary

- move the active LIBERO prompt content from Markdown into Python system and user prompt modules
- assemble prompt sections through `prompt_bundle.py` using the existing mapping renderer
- render workflow steps with `Numbered`
- keep the prompt content functionally consistent while moving cell context into the user prompt and removing the duplicated `TASK` section
- remove obsolete prompt files, stale documentation, and the unused memory sync helper

## Test

- compared the rendered prompts before and after the cleanup; instructions remain materially the same, with only expected section placement and whitespace differences
- verified all prompt template variables render without unresolved placeholders


robots/libero/prompts/system.py
```diff
@@ -1,193 +1,524 @@
...
+  xyz+pitch+yaw). It reaches several cm deeper.
+
+GRASPING:
+- **MUGS / BOWLS / CUPS grasp at the RIM, not the center**: SAM3/back-projection
+  give the object CENTER; closing the gripper there grabs air. Aim
+  `eef_y = object_y + 0.045` so Pi0 rim-hooks. (Mugs do NOT hang 4.5cm in -y like
+  bowls — you can wrist-segment the grasped object to measure the true held offset.)
+- Some objects grasp best with `pi0_pick` from the **DEFAULT HOME pose** (no
+  pre-position) — Pi0 has its own approach trajectory; pre-positioning can hurt.
+- **`pi0_pick` is reusable and repurposable** (only `pi0_end_to_end` is forbidden):
```
Sonorous281: Is the pi0_end_to_end a real tool?  The prompt says "only pi0_end_to_end is forbidden," but no tool by that name is registered

robots/libero/prompts/system.py
```diff
@@ -1,193 +1,524 @@
...
+   agentview; basket/cavity points are interior-centered; manipulation offsets are
+   planned from the perceived final_xyz. If this fails, keep perceiving — only
+   then start the manipulation plan. Re-verify with the newest image after every
+   command and update the table if anything moves.
+
+(xyz from agentview and wrist `back_project` are in the SAME world frame,
+directly comparable. Do NOT blindly average them — accept wrist coords only when
+consistent with the agentview anchor, or for basket/cavity geometry.)"""
+
+WORKFLOW_STEPS = (
```
qurakchin, line 390: the lines before are separated, but this attr is not.

robots/libero/prompt_bundle.py
```diff
@@ -1,35 +1,41 @@
...
-
-
-def user_prompt() -> dict[str, PromptNode]:
-    """Return the first user message tree."""
-    sections = dict(base_prompt.USER)
-    sections["Mode"] = libero_prompt.USER_MODE
-    return sections
+    """Assemble the LIBERO system prompt tree."""
+    return {
+        "Role and Evaluation": system_parts.ROLE_AND_EVALUATION,
```
qurakchin: Case sensitivity is not aligned with the prev .md

qurakchin (commented):
is the success rate tested?

Sonorous281:
After this change, no env uses the shared prompt blocks in rpent/context/prompts/prompt.py anymore, but add_robot.rst still tells new envs to use them.

robots/libero/prompts/system.py
wilburx813: fixed

wilburx813:
> is the success rate tested?

the prompt content is unchanged and only its formatting was refactored, so i believe it won't affect success rate

robots/libero/prompts/system.py
wilburx813: fixed

robots/libero/prompt_bundle.py
wilburx813: fixed

Sonorous281 (approved):

rpent/context/prompt_utils.py
qurakchin, line 100: add an example in function comment

qurakchin, line 64: add an example in function comment

wilburx813: Done

wilburx813: Done

qurakchin: add an example with a multi line items input. the `prefix = f"{index}. " if ordered else "- "` and `len(prefix)` is confusing.

qurakchin: add an example of "inspect title\nmulti-line inspect content"

qurakchin (approved):

qurakchin (approved):

PR #41, opened 2026-07-23
refactor(envs): move env-specific CLI + runtime init behind EnvSpec

qurakchin:
Introduce two Callable extension points on EnvSpec — parse_args and init_runtime — plus a RunConfig dataclass carrying the derived per-run identifiers (recipe_tag, output_dir, prompt_vars, dashboard_state, task_desc). rpent/cli/main.py is now env-agnostic: it no longer knows about LIBERO's suite/task/seed shape, endpoint parsing, or the env_server/vla_server subprocess wiring. Dropping a package under robots/ is enough to plug in a new env.


jx-qiu:
Nice, this is the long-awaited refactor for env interfaces. Let me check if the current EnvSpec abstraction is good enough for lerobot and franka.

rpent/envs/env_spec.py
```diff
@@ -6,18 +6,46 @@
...
 
+import argparse
+from collections.abc import Callable
 from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
 
 from rpent.envs.prompt_bundle import PromptBundle
 
+if TYPE_CHECKING:
```
jx-qiu, line 23: Why type checking?

jx-qiu (approved):

rpent/envs/env_spec.py
qurakchin: to avoid from recursive import.

andylin-hao (approved):

andylin-hao (approved):

PR #53, opened 2026-07-24
fix: fail fast when a spawned server crashes before ready

### LINKED ISSUE #45: RPent RPC daemon not detecting env server crashes
RPent gets stuck instead of printing a clear error message when the underlying env server dies.

For instance, RPent is stuck after printing `env_server spawned`:

```
2026-07-23 03:38:35 I [agent] physical agent cmd: /mnt/public/jxqiu/physicalagent/.venv/bin/python rpent/cli/main.py --env libero --suite libero_object_swap --task 2 --seed 0 --planner api --model anthropic:claude-opus-4-8 --max-tokens 8192
2026-07-23 03:38:37 I [daemon] env_server spawned (pid=2876296)
```

Meanwhile, the env server is already dead:

```
Traceback (most recent call last):
  File "/mnt/public/jxqiu/physicalagent/robots/libero/env_server.py", line 29, in <module>
    import torch  # noqa: E402
    ^^^^^^^^^^^^
ModuleNotFoundError: No module named 'torch'
```

qurakchin:
## Summary

Fixes #45.

`wait_for_ready` polled `healthz` for the full timeout even when the spawned `env_server` / `vla_server` had already crashed on startup (e.g. `ModuleNotFoundError: No module named 'torch'`), leaving the run hung after `env_server spawned`.

- Add `ProcessDaemon.poll()` exposing the subprocess exit code.
- `wait_for_ready` now watches the daemon and raises immediately with the exit code (and the tail of the crashed child's log) on early crash, instead of blocking until timeout.
- Also poll `healthz` on the endpoint-attach path. There is no local process to watch there, so an unreachable external server still waits out `timeout_s`, but it now ends with a clear `TimeoutError` rather than silently proceeding to fail on the first real RPC.
- Applied the same treatment to the SAM3 server spawn path.

## Test plan

- [x] Unit tests (`tests/test_wait_for_ready.py`): fail-fast on crashed daemon, log-tail surfaced in the error, and timeout when no daemon is given.
- [ ] Manual: spawn with a broken server (missing `torch`) and confirm a clear error appears within a poll interval instead of hanging.


rpent/utils/rpc.py
```diff
@@ -32,11 +35,29 @@ def wait_for_ready(
...
     while time.time() < deadline:
+        if daemon is not None:
+            rc = daemon.poll()
+            if rc is not None:
+                detail = last_err if last_err is not None else "no healthz attempt yet"
+                msg = (
+                    f"{daemon.name} exited with code {rc} before becoming "
+                    f"ready; check its log. last healthz error: {detail}"
+                )
+                tail = daemon.tail_log()
```
jx-qiu: It might be better if we remove the tail log from main logs and let the user directly read the server logs.

jx-qiu (approved):

tests/test_wait_for_ready.py
```diff
@@ -0,0 +1,46 @@
+"""Tests for wait_for_ready fail-fast / timeout behavior (stdlib unittest)."""
```
wilburx813: maybe we don't need to add this file to main?

wilburx813 (approved):

Sonorous281 (approved):

tests/test_wait_for_ready.py
jx-qiu: I removed the test. Tests are nice but I guess we need a principled way to approach that. Let's leave that as future work.

PR #54, opened 2026-07-24
fix: export segment calls in libero recipes

wilburx813:
## Summary

- include successful LIBERO segment artifacts when exporting recipe JSONL
- preserve prompt- and point-based segment inputs

## Motivation

Recipe export previously omitted segmentation calls, so replayed recipes could lose localization steps required by later robot actions.


robots/libero/tools.py
```diff
@@ -895,12 +894,34 @@ def write_recipe_from_states(output_dir: str, recipe_tag: str) -> str:
         result = entry.get("result")
         if isinstance(result, dict) and result.get("error"):
             continue
-        commands.append(command)
+        command_events.append(((step_idx, -1), command))
+
+    for artifact in (Path(output_dir) / "segments").glob("segment_*.json"):
```
qurakchin, line 899: why not add segment to primitive_actions?

wilburx813: state.json doesn't include segment tool calls.

Sonorous281:
The same description in strict_hybrid_guide.md:543 is now stale. It still says the recipe is
exported only "from the non-error primitive commands in states.json".

wilburx813:
> The same description in strict_hybrid_guide.md:543 is now stale. It still says the recipe is exported only "from the non-error primitive commands in states.json".

fixed

qurakchin (approved):

Sonorous281 (approved):

PR #58, opened 2026-07-27
fix: record Claude thinking blocks

wilburx813:
## Summary

Record non-empty Claude SDK `ThinkingBlock` content in transcripts and forward it to the dashboard.

Previously, these blocks were only preserved in the raw stream. Empty thinking blocks are ignored, and no dashboard changes are required.

## Test

- Ruff
- Recorder probe with a real `ThinkingBlock`


qurakchin (approved):
Does this problem exist in codex?

wilburx813:
> Does this problem exist in codex?

doesn't exist

Sonorous281 (approved):

PR #60, opened 2026-07-28
refactor(libero): remove unused primitives and centralize artifact paths

wilburx813:
## Summary

This PR cleans up the libero tool implementation.

- Remove unused primitive methods: `place` and `run_full_task`.
- Centralize artifact path handling across the libero tools and toolkit.

## Test

This change is limited to code cleanup and artifact handling; it does not alter task execution logic, so no impact on success rate is expected.


robots/libero/tools.py
```diff
@@ -15,11 +16,70 @@
...
+    "world_wrist",
+    "wrist_meta",
+    "images_wrist_hi",
+    "world_wrist_hi",
+    "segments",
+    "action_videos",
+)
+
+
+def artifact_path(
```
qurakchin, line 56: it's better to use args order (output_dir, kind, *, step, camera, resolution). move optional args behind.

qurakchin (commented):
This pr is big and i think a full task is needed

wilburx813:
> This pr is big and i think a full task is needed

for the full task tool, it directly calls the vla with task instruction and does not rely on the agent loop. So maybe is better to be a standalone test path instead of exposing it as a rpent tools, I think we can remove it for now and add it later if needed?

robots/libero/tools.py
wilburx813: fixed in [e218c2f](https://github.com/RLinf/RPent/pull/60/commits/e218c2f39aff0a83bd27142ff3476b85512521be)

qurakchin (approved):

robots/libero/tools.py
```diff
@@ -1720,67 +1605,44 @@ def view_driver_state(step: int | None = None) -> dict:
...
+            artifact_path(out_dir, "image", step=nn, camera=camera, resolution=resolution),
+            artifact_path(out_dir, "world", step=nn, camera=camera, resolution=resolution),
+        )
+        for resolution in ("high", "low")
+    ]
 
-def _select_segment_artifacts(nn: int, camera: str):
-    pairs = _segment_artifact_pairs(nn, camera)
     for image_path, world_path in pairs:
         if image_path.exists() and world_path.exists():
```
Sonorous281, line 1643: Could we keep the image-only fallback? RGB can be saved even when world-map generation fails; the old code still returned SAM results with world_error.

robots/libero/tools.py
```diff
@@ -649,62 +651,6 @@ def set_gripper(
...
-        }
-        for k, v in raw.items():
-            if k.endswith("_pos") and "robot0" not in k and "to_robot" not in k:
-                obj_name = k[:-4]
-                out["objects"][obj_name] = [float(x) for x in v]
-        return out
-
-    # ---- full task baseline (no instruction override) ----
-
-    def run_full_task(
```
Sonorous281, line 675: Could you update pro_hybrid_guide.md too. The agent is required to read it, but it still says run_full_task exists.

robots/libero/tools.py
```diff
@@ -831,36 +787,16 @@ def segment(
...
-    if len(arr) == idx:
-        arr.append(blob)
-    else:
-        arr[idx] = blob
-    with open(tmp, "w") as f:
-        json.dump(arr, f, indent=2)
-    os.replace(tmp, path)
+        states = []
+    states.append(blob)
+    with open(path, "w") as f:
```
Sonorous281: Could we preserve the previous tmp-file + os.replace() write? Direct "w" truncation can leave states.json corrupted on a failed write or concurrent read.

robots/libero/tools.py
wilburx813: i think we should make sure sam3 return complete result, return a world_error to planner seems meaningless, what do you think

Sonorous281: Could this behavior change affect task success when RGB is available but world-map generation fails? Previously, the planner could still use the SAM result for recovery, while the new code fails before running segmentation.

Sonorous281 (approved):

robots/libero/tools.py
wilburx813: I don’t think this will affect the success rate. because without world map the back_project will failure too, and this failure path should very rarely occur in practice.

PR #61, opened 2026-07-28
fix: separate termination and truncation

wilburx813:
## Summary

- Keep `terminated` and `truncated` separate in the LIBERO client.
- Make `libero_terminated` and primitive success fields reflect only official LIBERO termination.
- Record truncation explicitly and raise LIBERO episode-step defaults from 600 to 10000.

## Why

The client previously treated either `terminated` or `truncated` as the same `episode_done` state, and `dump_state()` exposed that combined state as `libero_terminated`. A time-limit truncation could therefore be reported as task success.


Sonorous281:
Could we document episode_truncated in the tool description and system prompt? The agent should stop when episode_truncated=true, because the episode has already ended and further actions will raise an error.

Sonorous281 (approved):

qurakchin (approved):

PR #62, opened 2026-07-29
fix: align EGL device with --cuda-device to prevent env_server crash on multi-GPU

### LINKED ISSUE #51: Bug: --cuda-device N crashes env_server on machines where EGL device order ≠ CUDA device order
--cuda-device or CUDA_VISIBLE_DEVICES != MUJOCO_EGL_DEVICE_ID

qurakchin:
## Summary

`--cuda-device N` sets `CUDA_VISIBLE_DEVICES` but MuJoCo's EGL backend still defaults to EGL device 0. When EGL enumeration order differs from CUDA order, `env_server` renders on the wrong GPU or a software/DRM-only EGL entry and crashes.

## Changes

- **`rpent/utils/egl.py`** (new) — bridges CUDA and EGL enumerations via `EGL_NV_device_cuda`; `configure_egl_device()` sets `MUJOCO_EGL_DEVICE_ID`
- **`robots/libero/env_server.py`** — defers heavy imports, calls `configure_egl_device()` after parsing `--cuda-device`
- **`robots/libero/vla_server.py`** — adds `--cuda-device` for `CUDA_VISIBLE_DEVICES` only (no EGL)
- **`robots/libero/__init__.py`** — passes `--cuda-device` to subprocesses via command line
- **Docs** — `advanced_deployment.rst` examples use `--cuda-device` instead of `export CUDA_VISIBLE_DEVICES`

Fixes #51.

### DIFF
--- a/robots/libero/env_server.py
+++ b/robots/libero/env_server.py
@@ -26,10 +26,11 @@ if str(RLINF_REPO_PATH) not in sys.path:
 os.environ.setdefault("ROBOT_PLATFORM", "LIBERO")
 
 import numpy as np
-import torch
 from omegaconf import OmegaConf
 
 from rlinf.envs.libero.libero_env import LiberoEnv
+
+# torch import is deferred into main() after --cuda-device sets CUDA_VISIBLE_DEVICES.
 
 
 # ---------------------------------------------------------------------------
@@ -298,9 +299,20 @@ def main():
     args = p.parse_args()
 
     if args.cuda_device is not None:
-        os.environ["CUDA_VISIBLE_DEVICES"] = str(args.cuda_device)
+        assert "torch" not in sys.modules, \
+            "torch must not be imported before --cuda-device is applied"
+        target = str(args.cuda_device)
+        prev = os.environ.get("CUDA_VISIBLE_DEVICES")
+        if prev is not None and prev != target:
+            logger.warning(
+                "CUDA_VISIBLE_DEVICES=%s is already set; overriding with --cuda-device=%s",
+                prev, args.cuda_device,
+            )
+        os.environ["CUDA_VISIBLE_DEVICES"] = target
         from rpent.utils.egl import configure_egl_device
         configure_egl_device(args.cuda_device)
+
+    import torch
 
     raw_env = make_env(args.task, args.seed, suite_name=args.suite,
                        max_episode_steps=args.max_episode_steps)
--- a/robots/libero/env_server.py
+++ b/robots/libero/env_server.py
@@ -302,22 +302,34 @@ def main():
     p.add_argument("--parent-watch", action="store_true",
                    help="watch parent process via stdin pipe and exit when it dies")
     p.add_argument("--cuda-device", type=int, default=None,
-                   help="GPU device exposed through CUDA_VISIBLE_DEVICES.")
+                   help="GPU device to pin MuJoCo EGL rendering and the torch "
+                        "default device to (physical CUDA ordinal).")
     args = p.parse_args()
 
     if args.cuda_device is not None:
-        assert "torch" not in sys.modules, \
-            "torch must not be imported before --cuda-device is applied"
-        target = str(args.cuda_device)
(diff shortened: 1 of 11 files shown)


wilburx813 (commented):
it seems we don't support export CUDA_VISIBLE_DEVICES now? could we use the approach in https://github.com/google-deepmind/mujoco/pull/3246 ？

qurakchin:
> it seems we don't support export CUDA_VISIBLE_DEVICES now? could we use the approach in [google-deepmind/mujoco#3246](https://github.com/google-deepmind/mujoco/pull/3246) ？

CUDA_VISIBLE_DEVICES is still supported, but the EGL device cannot be aligned to cuda device. the approach in mujoco is similar to this approach, only the usage is different. I will assess it.

wilburx813 (approved):

qurakchin:
> > it seems we don't support export CUDA_VISIBLE_DEVICES now? could we use the approach in [google-deepmind/mujoco#3246](https://github.com/google-deepmind/mujoco/pull/3246) ？
> 
> CUDA_VISIBLE_DEVICES is still supported, but the EGL device cannot be aligned to cuda device. the approach in mujoco is similar to this approach, only the usage is different. I will assess it.

It's quite complicated. I think the solution now is simple enough.

rpent/utils/egl.py
```diff
@@ -0,0 +1,93 @@
...
+        return
+    cuda_ordinal = cuda_device
+
+    try:
+        mapping = cuda_to_egl_map()
+    except Exception as e:
+        logger.warning("EGL device probe failed (%s); leaving MUJOCO_EGL_DEVICE_ID unset", e)
+        return
+
+    egl_index = mapping.get(cuda_ordinal)
```
Sonorous281, line 81: Could we support both the physical CUDA ID and the logical ID after CUDA_VISIBLE_DEVICES remapping? For example, physical GPU 2 may appear as logical GPU 0, causing mapping.get(2) to fail even when the correct EGL device exists.

robots/libero/env_server.py
```diff
@@ -25,11 +25,11 @@
     sys.path.insert(0, str(RLINF_REPO_PATH))
 os.environ.setdefault("ROBOT_PLATFORM", "LIBERO")
 
-import numpy as np  # noqa: E402
-import torch  # noqa: E402
-from omegaconf import OmegaConf  # noqa: E402
+import numpy as np
+import torch
+from omegaconf import OmegaConf
```
Sonorous281: Could we set CUDA_VISIBLE_DEVICES before importing torch and LiberoEnv, or preserve it in the subprocess environment? These imports may initialize CUDA or MuJoCo before --cuda-device is applied.

jx-qiu:
> > it seems we don't support export CUDA_VISIBLE_DEVICES now? could we use the approach in [google-deepmind/mujoco#3246](https://github.com/google-deepmind/mujoco/pull/3246) ？
> 
> CUDA_VISIBLE_DEVICES is still supported, but the EGL device cannot be aligned to cuda device. the approach in mujoco is similar to this approach, only the usage is different. I will assess it.

If my understanding is correct, with your PR, RPent respects both `CUDA_VISIBLE_DEVICES` and `--cuda-device`, and `--cuda-device` overrides `CUDA_VISIBLE_DEVICES`. Setting either option will make all GPU work, regardless of rendering or cuda compute, performed by the specified GPU. Is that right?

rpent/utils/egl.py
qurakchin: It's complicate to use both CUDA_VISIBLE_DEVICES and --cuda-device. I think only one is enough.

Sonorous281: Using --cuda-device as the single recommended mechanism is fine, but CUDA_VISIBLE_DEVICES has not actually been disabled: _subprocess_env() still inherits it when --cuda-device is omitted. Could we either handle that path correctly or reject it explicitly?

Sonorous281 (approved):

PR #63, opened 2026-07-30
refactor: consolidate RPC envelope, error handling, and dispatch into shared layer

qurakchin:
## Summary
- Eliminated copy-pasted `dispatch`/dispatch-lock/error-construction between `HttpRpcServer` and `SocketRpcServer`
- Centralized response envelope (`make_error_response` / `check_response`) into `rpent/utils/rpc.py`
- Consolidated `healthz`/`shutdown`/dispatch-lock into a single closure in `RpcFacade.serve()`
- Removed dead/useless `req_id` (UUID) and `close()` from both transports
- Moved `parse_endpoint` from `robots/libero/__init__.py` to `rpent/utils/rpc.py`
- Fixed missing `_from_json` call in `HttpRpcClient.call` causing numpy arrays to not be deserialized on the client side

## Test plan
- [ ] Import tests pass
- [ ] HTTP transport integration smoke test passes
- [ ] Socket transport integration smoke test passes
- [ ] LIBERO robot scenario runs end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/robots/libero/__init__.py
+++ b/robots/libero/__init__.py
@@ -136,21 +136,6 @@ def _parse_config(args: argparse.Namespace) -> RunConfig:
     )
 
 
-def _parse_endpoint(endpoint: str) -> tuple[str, str, int]:
-    """Parse ``[protocol://]host:port`` into ``(protocol, host, port)``.
-
-    Protocol defaults to ``http`` when the prefix is omitted.
-    """
-    if "://" in endpoint:
-        protocol, _, rest = endpoint.partition("://")
-    else:
-        protocol, rest = "http", endpoint
-    host, _, port = rest.partition(":")
-    if not host or not port:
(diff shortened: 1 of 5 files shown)


qurakchin:
the core fix is the `response = _from_json(json.loads(raw))` line in http_rpc. others are refactors.

wilburx813 (approved):

Sonorous281:
Could shutdown also acquire _lock? Otherwise it may trigger server teardown before an in-flight long RPC finishes, changing the previous shutdown behavior.

qurakchin:
> Could shutdown also acquire _lock? Otherwise it may trigger server teardown before an in-flight long RPC finishes, changing the previous shutdown behavior.

fixed

Sonorous281 (approved):

PR #66, opened 2026-08-03
docs: fix cross-reference labels, typos, and backtick errors

qurakchin:
## Summary

- Fix cross-reference labels and confval/ref role names in RST docs
- Add missing backticks and fix typos across docs
- Fix primitives's -> primitives' grammar
- Fix get_output_dir() return type with assertion guard
- Add BrokenPipeError/ConnectionResetError guard in HttpRpcServer.do_POST
- Add get_output_dir() default to list_dir tool
- Tighten HttpMcpServer lifespan loop
- Add error/task_description state tracking in dashboard
- Remove stale robocasa-specific content from generic docs
- Sync Chinese translations with English docs

### DIFF
--- a/rpent/planner/utils/http_mcp_server.py
+++ b/rpent/planner/utils/http_mcp_server.py
@@ -18,8 +18,6 @@ Usage::
 from __future__ import annotations
 
 import asyncio
-import json
-import logging
 import socket
 import threading
 from typing import Any
@@ -32,8 +30,9 @@ from mcp.server.lowlevel import Server
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 
 from rpent.tools.toolkit import Toolkit
+from rpent.utils.logging import get_logger
 
-logger = logging.getLogger("mcp_http")
+logger = get_logger("mcp_http")
 
 SERVER_NAME = "rpent"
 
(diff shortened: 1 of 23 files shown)


jx-qiu (approved):

wilburx813 (approved):

rpent/planner/utils/http_mcp_server.py
```diff
@@ -214,5 +220,11 @@ def stop(self, *, timeout_s: float = 5.0) -> None:
             self._server.should_exit = True
         if self._thread is not None:
             self._thread.join(timeout=timeout_s)
+            if self._thread.is_alive():
```
Sonorous281, line 223: Despite the warning saying that references are left intact, the following assignments discard them, so the caller can no longer retry shutdown or detect the leaked thread.

rpent/dashboard/state.py
```diff
@@ -112,6 +112,14 @@ def mark_done(self, terminated: bool | None = None) -> None:
                 terminated = any(item.get("terminated") for item in self._timeline)
             self._terminated = bool(terminated)
 
+    def mark_failed(self, error: str | None = None) -> None:
```
Sonorous281: failed is now a terminal state, but related responses only expose the episode video when state is done. This hides existing videos for failed runs, which are useful for debugging. Shall we treat both done and failed as terminal when checking video availability?

rpent/planner/utils/http_mcp_server.py
qurakchin: It's in the stop process so the leaked thread will not cause heavy influence.

PR #73, opened 2026-08-06
feat(env): manage env state centrally

jx-qiu:
This pull request updates the documentation to reflect recent changes in how environment state, tool registration, and artifact management are handled in the codebase. The main focus is on the transition from manual tool registration and file management to an automated, `EnvState`-centric workflow. Documentation in both English and Chinese is revised for clarity and accuracy, and references to legacy interfaces and files are replaced with the new conventions.

**Major documentation and interface updates:**

*General workflow and tool registration:*
- Updated instructions for adding scripted primitives: tool registration is now automatic based on `TOOLS_SPEC` and `@readonly`. No need to manually call `add_tools` anymore! [[1]](diffhunk://#diff-76aa956b723bab0b2f7b4d91ccf435e3018d1117a95bf58566ce15f293d59d77L38-R64) [[2]](diffhunk://#diff-76aa956b723bab0b2f7b4d91ccf435e3018d1117a95bf58566ce15f293d59d77L70-R82) [[3]](diffhunk://#diff-70ca6faf9f73d6fb145b996b49600013fb1cd5e17d312e28f4004c1aa85a086bL36-R60) [[4]](diffhunk://#diff-70ca6faf9f73d6fb145b996b49600013fb1cd5e17d312e28f4004c1aa85a086bL66-R77)
- Clarified that tool return values should be small dicts, and large observations must be saved via `EnvState.save`; images and geometry are exposed through dedicated tools, not raw paths. [[1]](diffhunk://#diff-76aa956b723bab0b2f7b4d91ccf435e3018d1117a95bf58566ce15f293d59d77L170-R179) [[2]](diffhunk://#diff-70ca6faf9f73d6fb145b996b49600013fb1cd5e17d312e28f4004c1aa85a086bL155-R162)

*State management and artifact handling:*
- Revised descriptions of per-step state dumps: now handled by `EnvState.record_step` and `EnvState.save`, with explicit artifact tracking. Artifacts are referenced by logical base names, not manual paths. [[1]](diffhunk://#diff-ce0e2ce7692aae4dc7bb52a8efb5b8e4da380e2f99a9dd47ce8978a55330fdcbL223-R245) [[2]](diffhunk://#diff-5545e8429be10fe260731daecb2691f7ed2e8098fc1b7a64aaf2b5a5ffae3bacL210-R230)
- Updated conventions to emphasize that all environment observations are owned and managed by `EnvState`, and that callers must not construct storage paths directly. [[1]](diffhunk://#diff-ce0e2ce7692aae4dc7bb52a8efb5b8e4da380e2f99a9dd47ce8978a55330fdcbL249-R263) [[2]](diffhunk://#diff-5545e8429be10fe260731daecb2691f7ed2e8098fc1b7a64aaf2b5a5ffae3bacL235-R247)

*Tool interface and naming consistency:*
- Standardized on `view_env_state` instead of `view_driver_state` throughout the documentation and interface references. [[1]](diffhunk://#diff-4447f6b3a8284227164ef51e2051a84375f960e12a757f40e03eb4c9ed4fb19fL93-R93) [[2]](diffhunk://#diff-ed4ec0eda326a95dfd056c9d4efadd7e3b8cdc565be4a292d82915d914fe82d2L119-R119) [[3]](diffhunk://#diff-ed4ec0eda326a95dfd056c9d4efadd7e3b8cdc565be4a292d82915d914fe82d2L150-R153) [[4]](diffhunk://#diff-eb50c0ae2154f14d641fea33ec6e2dcf5e33ee986c22de732cc75b610a6c8f31L90-R90) [[5]](diffhunk://#diff-8909ad6e6fe5a510d2295afe05432ce3ac00d43e24fa34f54ff9828ff25d74bbL115-R115) [[6]](diffhunk://#diff-8909ad6e6fe5a510d2295afe05432ce3ac00d43e24fa34f54ff9828ff25d74bbL144-R147)

*Quickstart and calibration documentation:*
- Updated quickstart guides to explain the new artifact naming, storage, and retrieval conventions, including the use of `view_env_state(step=-1)` for inspecting the latest state. [[1]](diffhunk://#diff-deb89cc708fabbcb50246d5702c022895d39a7f8bf8ed5261a6e28cd14ac4c1bL106-R111) [[2]](diffhunk://#diff-c17bb93cc804f83e2d10c7dc3351bc48e555386e8d12f82e9a868c0ad5d42b1aL98-R102)
- Improved calibration guide instructions to use `view_env_state({"step": 0})` for initial state and clarified that direct file or BDDL access is not supported. [[1]](diffhunk://#diff-2c413b0b2d21bae79f4d3286b764a452b2d5f21662dde3a5e464820ef5c23267L5-R13) [[2]](diffhunk://#diff-2c413b0b2d21bae79f4d3286b764a452b2d5f21662dde3a5e464820ef5c23267L20-R25)

*Code cleanup:*
- Removed the unused `video_path` parameter from `get_toolkit` in `robots/libero/__init__.py`.


rpent/tools/toolkit.py
```diff
@@ -150,39 +194,103 @@ def _register_common_tools(self) -> None:
...
                 )
             operation = _ToolOperation()
             self._active_operation = operation
 
         try:
+            started = time.perf_counter()
+            failed = False
             try:
                 result = handler(**input_dict)
             except TypeError as e:
```
Sonorous281, line 218: Could this TypeError path bypass get_env_state()? If a stateful primitive mutates the env and then raises TypeError, the planner may see stale state. Should this path also go through post-action state capture?

robots/libero/tools.py
```diff
@@ -1099,86 +1040,35 @@ def dump_state(primitives: LiberoPrimitives, output_dir: str, step_idx: int,
...
             world_wrist_hi,
-        )
-        wrist_world_map_hi = _artifact_relative_path(
-            step_idx, "wrist", "high", "world"
+            step=step_idx,
         )
     except Exception as e:
         logger.warning("wrist high-res dump failed: %s", e)
 
-    for old_step in range(max(0, int(step_idx) - 4)):
```
Sonorous281, line 1120: Could we keep the previous retention policy for high-res artifacts here? Without cleanup, the two 1024×1024 world maps alone add 12 MB per step, which may become significant for long or batched rollouts.

rpent/tools/toolkit.py
jx-qiu: I think TypeError in this case can only mean that the planner passes the wrong number/type of arguments to the primitive, in which case the exception throws before the control reaches the env server.

rpent/tools/toolkit.py
```diff
@@ -131,9 +174,10 @@ def add_tool(
             spec: Anthropic-shaped tool schema dict (``name``,
                 ``description``, ``input_schema``).
             handler: Callable invoked with the tool's input kwargs; returns
-                a result dict.
+                a result dict. Decorate state-advancing primitives with
+                :func:`updatestate`.
         """
-        self._tools[name] = (spec, handler)
+        self._tools[name] = (spec, handler, _updates_state(handler))
```
Sonorous281: Would it be safer to make stateful and read-only behavior more explicit? Missing @updatestate currently seems to silently treat a mutating primitive as read-only, which could leave stale observations.

rpent/tools/state.py
```diff
@@ -0,0 +1,465 @@
...
+        try:
+            with temporary.open("w") as file:
+                json.dump(manifest, file, indent=2, default=_json_default)
+            os.replace(temporary, destination)
+        finally:
+            temporary.unlink(missing_ok=True)
+
+    # -- lifecycle and counters -----------------------------------------
+
+    def reset(self) -> None:
```
Sonorous281, line 138: Could this leave stale artifacts when EnvState is recreated for an existing output directory? At initialization _steps and _run_artifacts are empty, so reset() removes the old manifest but doesn't seem to know which artifacts from the previous process should be deleted.

jx-qiu: Good catch. Will fix it.

rpent/tools/toolkit.py
jx-qiu: It might be better to reverse the @updatestate annotation semantics, for instance using @readonly to mark primitives that are read-only, and update state by default. Let me fix it.

rpent/tools/toolkit.py
Sonorous281: Could TypeError also be raised inside the primitive after it has already executed env.step()? In that case the env may have changed, but this path would return without updating the state.

rpent/tools/toolkit.py
jx-qiu: Fixed in https://github.com/RLinf/RPent/pull/73/commits/27659541b5622ba3b807b22d7f34f66295d5aaf7.

rpent/tools/state.py
jx-qiu: Fixed in https://github.com/RLinf/RPent/pull/73/commits/c70469e9943f1abd5ea4d653b01678eb9931a3a5.

rpent/tools/state.py
```diff
@@ -0,0 +1,472 @@
...
+        """Remove all artifacts and start a fresh trace."""
+        self._output_dir.mkdir(parents=True, exist_ok=True)
+        for path in self._output_dir.iterdir():
+            if path.is_file() and path.suffix.lower() != ".log":
+                path.unlink()
```
wilburx813: reset every file expect log seems unsafe, the user may accidentally pass a important directory

rpent/tools/state.py
```diff
@@ -0,0 +1,472 @@
...
+            elif suffix == ".bin":
+                temporary.write_bytes(bytes(value))
+            else:
+                raise ValueError(f"unsupported artifact suffix: {suffix}")
+            os.replace(temporary, destination)
+            if record is not None:
+                record.artifacts.add(name)
+            else:
+                self._run_artifacts.add(name)
+            self._write_manifest()
```
wilburx813: could we write manifest once per step instead of rewriting it after every artifact save?

rpent/tools/toolkit.py
jx-qiu: Fixed in https://github.com/RLinf/RPent/pull/73/commits/6b0e115f9f0644d78a3042cccfdd813d6f4de721.

wilburx813:
btw, seems rpent-memory needs update too.

robots/libero/tools.py
jx-qiu: For libero, each step produces ~17 MiB of on-disk artifacts. I will see if there are opportunities for better lossless compression.

jx-qiu:
> btw, seems rpent-memory needs update too.

Do you mean the markdown files under resources/libero/memory?

wilburx813:
> > btw, seems rpent-memory needs update too.
> 
> Do you mean the markdown files under resources/libero/memory?

Yes, some files under resources/libero/memory still reference something like view_driver_state

qurakchin (approved):

rpent/tools/state.py
jx-qiu: Right. https://github.com/RLinf/RPent/pull/73/commits/e85221f1a194c149770ff6853d3e0b4a28d740b8 reverts the behavior; all the artifacts are preserved.

wilburx813 (approved):

jx-qiu:
> > > btw, seems rpent-memory needs update too.
> > 
> > 
> > Do you mean the markdown files under resources/libero/memory?
> 
> Yes, some files under resources/libero/memory still reference something like view_driver_state

We need versioning mechanisms with the rpent-memory HF repo. @Sonorous281 and I will discuss how to. For now we have to keep the old memory files because the "reproduce/libero" branch needs it (to reproduce HarnessVLA results).

rpent/tools/state.py
jx-qiu: Nice suggestion. Implemented in https://github.com/RLinf/RPent/pull/73/commits/773bb3bf04ba4123834065be0700647e6c96708a.

robots/libero/tools.py
jx-qiu: I have added fast numpy compression in https://github.com/RLinf/RPent/pull/73/commits/97a66f016d0161f5d52ffa50b8eca16c5f02dee3 / https://github.com/RLinf/RPent/pull/73/commits/513ee09b3faf7d7663aed29f348eed409083dad6 and all depth and world maps are now compressed by default. We will need a in-memory cache mechanism for EnvState before fully implementing the previous retention policy.

jx-qiu: Compression is quite effective: the agentview_world_high world map shrinks from 6 MiB -> 2 MiB, in ~100 ms. Each step now produces 17 MiB -> 12 MiB artifacts.

Sonorous281 (approved):

PR #74, opened 2026-08-07
fix(api): handle read_image file errors

wkzCode:
## Summary
- resolve relative image paths from the repository root, matching `read_text_file`
- return structured tool errors for missing files, directories, and image read failures instead of aborting the API agent run
- preserve `ToolReturn` with binary image content for successful reads
- add coverage for missing, directory, read-failure, and successful relative-path cases

## Motivation
API planner models can occasionally mistype a long image path. `BinaryContent.from_path()` currently lets `FileNotFoundError` escape, which terminates the entire agent after earlier physical actions may already have executed. Returning the error to the model makes this behavior consistent with `read_text_file` and allows the model to recover safely.

## Validation
- `git diff --check`
- `python -m compileall -q rpent/planner/api_loop.py tests/test_api_read_image.py`
- pytest was not run locally because the available Python environments do not have `pydantic_ai`/pytest installed

### DIFF
--- a/rpent/planner/api_loop.py
+++ b/rpent/planner/api_loop.py
@@ -15,6 +15,7 @@ import dataclasses
 import json
 import queue
 from collections import deque
+from pathlib import Path
 from typing import Any
 
 from pydantic_ai import Agent, BinaryContent, ModelSettings, Tool, ToolReturn
@@ -43,6 +44,7 @@ from rpent.dashboard.interaction import DashboardInteractionPort, DashboardMessa
 from rpent.dashboard.planner_control import DashboardPlannerControl
 from rpent.planner.base import PlannerResult
 from rpent.tools.toolkit import Toolkit
+from rpent.utils.config import get_repo_root
 from rpent.utils.logging import get_logger
 
 logger = get_logger("api_loop")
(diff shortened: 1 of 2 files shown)


rpent/planner/api_loop.py
```diff
@@ -706,11 +708,27 @@ def _build_tools(toolkit: Toolkit, *, no_images: bool = False) -> list[Tool]:
...
+    matching :func:`rpent.tools.common.read_text_file`, so a bad model-supplied
+    path does not abort the entire agent run.
+    """
+    image_path = Path(path)
+    if not image_path.is_absolute():
+        image_path = get_repo_root() / image_path
+    if not image_path.exists():
+        return {"error": f"file not found: {image_path}"}
+    if image_path.is_dir():
+        return {"error": f"is a directory: {image_path}"}
```
jx-qiu, line 724: Since we already try catch BinaryContent.from_path(image_path), can we remove these lines and directly hand over path checking responsibility to that?

rpent/planner/api_loop.py
```diff
@@ -706,11 +708,27 @@ def _build_tools(toolkit: Toolkit, *, no_images: bool = False) -> list[Tool]:
...
+def read_image(path: str) -> ToolReturn | dict[str, str]:
+    """Read a local image path returned by an RPent tool as visual input.
+
+    File-system failures are returned to the model as structured tool errors,
+    matching :func:`rpent.tools.common.read_text_file`, so a bad model-supplied
+    path does not abort the entire agent run.
+    """
+    image_path = Path(path)
+    if not image_path.is_absolute():
+        image_path = get_repo_root() / image_path
```
jx-qiu, line 720: I suggest using `get_output_dir` instead of `get_repo_root` since run-time images are stored in the output directory, and by giving a relative url to an image, planners should want to access one generated at run-time?

jx-qiu (commented):
Thanks for the fix! Some minor comments:

rpent/planner/api_loop.py
jx-qiu: On second thoughts it's OK to leave that be for now. After #73 landed we can make the read_image tool use the new EnvState APIs instead of supplying raw paths.

wkzCode: > On second thoughts it's OK to leave that be for now. After #73 landed we can make the read_image tool use the new EnvState APIs instead of supplying raw paths.

That's more reasonable. Thanks for your great work!

jx-qiu (approved):

Sonorous281 (approved):

PR #76, opened 2026-08-08
refactor(libero): minimize SAM3 visual review

scilwb:
## Summary

- Add a one-shot visual review after text-prompted SAM3 segmentation.
- Reuse the model selected for the current eval across the `api`, `codex`,
  and `claude_code` backends; no task-specific LLM is hard-coded.
- Skip review when image input is disabled or the backend explicitly rejects
  visual input.
- Accept structured `KEEP`, `WRONG`, or `SUBTRACT` decisions.
- Allow `SUBTRACT` to remove pixels only after target-component, protected-core,
  subset, confidence, and world-projection checks pass.

## Motivation

SAM3 can occasionally merge the requested object with nearby distractors.
This change adds a conservative, subtraction-only visual review step without
changing the SAM3 service or allowing the reviewer to add pixels.

## Behavior

- `KEEP` and `WRONG` preserve the original SAM3 mask.
- Unsupported vision, timeout, malformed output, low confidence, or failed
  geometry/depth validation preserves the original SAM3 mask.
- `SUBTRACT` can only produce a strict subset of the raw mask.
- The retained target component must dominate the candidate mask.
- At least 90% of the protected target core must remain.
- World projection uses the validated target component rather than unrelated
  disconnected fragments.
- Point-prompt segmentation is unchanged.

## Scope

The final diff is limited to:

- `robots/libero/__init__.py`: inject the eval model configuration.
- `robots/libero/sam3_review.py`: backend adapters, structured review, and
  subtraction safety checks.
- `robots/libero/tools.py`: invoke review and fall back to raw SAM3 on failure.

No environment, VLA, SAM3 server, planner, dependency, or training code is
changed.

## Validation

- Python compilation passed for all changed production modules.
- Ruff checks passed.
- `git diff --check` passed.
- Existing repository test suite: `6 passed`.
- The final diff contains no fixed `gpt-5.5` or `xhigh` configuration.

## Known limitations

- Each reviewed text segmentation adds one model request, latency, and cost.


Sonorous281:
Could we rebase this onto the latest main first and port the review through EnvState? segment() has already moved away from path-based artifact access, while this patch still uses the old _load_step flow.

lwbscu:
Closing this for now — Chunyang will further refine it. Thanks for the feedback!

PR #79, opened 2026-08-10
feat(libero): add exploration and layered memory evaluation

Elessar123:
## Summary

- add multi-attempt LIBERO exploration with independent planner sessions
- add layered global/suite/task memory generation and evaluation
- automatically validate and merge successful exploration artifacts
- preserve the original Hugging Face memory and prompt as the default
- support API, Claude Code, and Codex planners
- document exploration and layered evaluation in English and Chinese

## Testing

- `pytest -q` — 16 passed


docs/source-en/rst_source/usage/libero.rst
```diff
@@ -100,6 +100,50 @@ Minimal command
...
+   rpent --env libero --suite libero_10_task --task 0 --seed 1 \
+     --planner claude_code --memory-profile hf
+
+Use ``layered`` to evaluate against the local global/suite/task corpus without
+overwriting it from Hugging Face:
+
+.. code-block:: bash
+
+   rpent --env libero --suite libero_10_task --task 0 --seed 1 \
+     --planner codex --memory-profile layered
```
QuanluZhang: how about using the name "local" instead of "layered"? according to my understanding, "layered" means using local existing memory?

docs/source-en/rst_source/usage/libero.rst
```diff
@@ -100,6 +100,50 @@ Minimal command
...
+
+.. code-block:: bash
+
+   rpent --env libero --suite libero_10_task --task 0 --seed 0 \
+     --planner api --model anthropic:claude-opus-4-8 \
+     --explore --explore-sessions 3 --explore-attempts-per-session 5 \
+     --memory-dir /path/to/local/libero-memory
+
+Pass ``--no-auto-merge-memory`` to retain inbox drafts for manual review.  The
+same publication implementation is available to maintainers directly:
```
QuanluZhang: what is the meaning of this sentence: "The
same publication implementation is available to maintainers directly"?

rpent/memory/cli.py
```diff
@@ -0,0 +1,54 @@
+"""Command-line maintenance interface for layered memory."""
```
jx-qiu: Let's rename and move this to rpent/cli/memory.py since it is a standalone cli entry point?

robots/libero/tools.py
```diff
@@ -1166,6 +1197,26 @@ def dump_state(primitives: LiberoPrimitives, output_dir: str, step_idx: int,
 )
 
 TOOLS_SPEC = [
+    {
```
jx-qiu, line 1100: Is the explore mode designed for libero only? The memory structure looks libero-specific. If there are no plans of porting explore mode to Robocasa / Robotwin, I suggest we (1) move the reset tool to libero, and optionally (2) only add the tool to LiberoToolkit when explore mode is enabled.

rpent/tools/toolkit.py
```diff
@@ -206,6 +206,18 @@ def raise_if_cancelled(self) -> None:
     def close(self) -> None:
         """Release the env-side primitives / servers at end of run. Default: no-op."""
 
+    def begin_session(self) -> None:
```
jx-qiu: I am still thinking about Toolkit lifecycles. Between two exploration attempts, should we just reuse an existing Toolkit instance (existing code), or release and reconstruct one?

docs/source-en/rst_source/usage/libero.rst
Elessar123: Thanks for the suggestion. I have renamed the CLI value from `layered` to `local` and updated the English and Chinese documentation accordingly.

docs/source-en/rst_source/usage/libero.rst
Elessar123: I have replaced it with a direct explanation that maintainers can use `rpent-memory` to validate the corpus, rebuild its index, or explicitly merge a reviewed inbox cell.

rpent/memory/cli.py
Elessar123: I have moved the standalone entry point from `rpent/memory/cli.py` to `rpent/cli/memory.py` and updated the `rpent-memory` entry in `pyproject.toml` to point to `rpent.cli.memory:main`.

robots/libero/tools.py
Elessar123: Yes, exploration is currently designed for LIBERO (but the same structure can be used for Robocasa 365, I will test it), and the reset behavior is LIBERO-specific. The reset tool schema and implementation are kept under `robots/libero`, and `LiberoToolkit` only registers the tool when `explore=True`.

rpent/tools/toolkit.py
Elessar123: I think reusing one is OK? What do you think?

robots/libero/prompts/explore.py
```diff
@@ -0,0 +1,437 @@
...
+memory gives the reasoning and failure modes needed to adapt them. In your
+audit, RECORD the memory files you read (or state that none matched), so memory
+consultation is auditable."""
+
+STEP_PRIMITIVES = """ALLOWED PRIMITIVES (physics-only; full schemas in the tool list/guides):
+`move_to`, `pi0_pick`, `pi0_doubled`, `release`, `set_gripper`,
+`rotate_wrist`, `rotate_pitch`, `move_pose`, AND `reset` (🔁 allowed here —
+close out the attempt first, see below).
+FORBIDDEN: `exit`, `set_object_pose`, `articulate_to`, `js_move_to`,
+`carry_object`.
```
QuanluZhang, line 196: are these primitives already written in libero code? There might be inconsistency if we duplicate the primitives twice. @qurakchin

robots/libero/__init__.py
```diff
@@ -72,6 +77,33 @@ def _add_cli_args(parser: argparse.ArgumentParser, use_dashboard: bool) -> None:
...
+        "--explore-sessions",
+        type=int,
+        default=3,
+        help="Independent planner sessions per exploration run (default: 3).",
+    )
+    parser.add_argument(
+        "--explore",
+        action="store_true",
+        help="Enable multi-attempt exploration and memory distillation.",
+    )
```
QuanluZhang: make these format consistent with the rest arguments

robots/libero/tools.py
QuanluZhang: we should think about how to make the memory more general, though not in this pr

rpent/cli/main.py
```diff
@@ -154,6 +172,8 @@ def main() -> int:
     args = parser.parse_args()
     if args.dashboard and args.interactive:
         parser.error("--dashboard and --interactive cannot be used together")
+    if args.dashboard and getattr(args, "explore", False):
+        parser.error("--dashboard and --explore cannot be used together")
```
QuanluZhang: why? can't I watch the whole exploration process in dashboard?

rpent/cli/main.py
```diff
@@ -240,27 +267,78 @@ def main() -> int:
...
+                system_prompt = prompt_bundle.render(
+                    "system",
+                    variables={
+                        **prompt_vars,
+                        "session_no": session_no,
+                        "session_max": sessions,
+                    },
+                )
+                session_msg = _handoff_message(output_dir, session_no, sessions)
+                toolkit.begin_session()
```
QuanluZhang: better to wrap this part into a function

rpent/planner/claude_code.py
```diff
@@ -808,6 +808,12 @@ def _tool_result_to_mcp(tr: Any) -> dict[str, Any]:
             )
 
     response: dict[str, Any] = {"content": content}
+    # Surface toolkit-level failures as MCP errors. Without this every result
+    # looks successful to the SDK, and `_Recorder` promotes a `finish` call to
+    # the run's finish_result even when the handler rejected it.
+    result = getattr(tr, "result", None)
+    if isinstance(result, dict) and result.get("error"):
+        response["isError"] = True
```
QuanluZhang: are these three lines duplicated with line 817-819?

QuanluZhang: only the keys are different?

rpent/tools/toolkit.py
QuanluZhang: what is your opinion? @qurakchin @Sonorous281

README.md
```diff
@@ -137,6 +137,30 @@ rpent --env libero --suite libero_object_swap --task 2 --seed 0 \
...
 
+### Exploration and local-memory evaluation
+
+Evaluation remains the default. Add `--memory-profile local` to evaluate
+against a local global/suite/task memory corpus:
+
+```bash
+rpent --env libero --suite libero_10_task --task 0 --seed 1 \
+  --planner codex --memory-profile local \
+  --memory-dir /path/to/libero-memory
```
QuanluZhang, line 148: I don't understand this part. what is different when I add `--memory-profile local`? I don't have the memory corpus locally at this point, right? Will this command explore? or should I downloaded the memory corpus from huggingface before running this command?

rpent/tools/toolkit.py
jx-qiu: Let me see. In cli/main.py, the planner is already reconstructed every exploration attempt. It might be cleaner to reconstruct a toolkit as well. This way we don't need to track session state inside the toolkit (a responsibility it is not supposed to carry). 

To implement toolkit reconstruction:
The toolkit is mostly stateless so we should be able to cleanly close and reconstruct one, except for the state files. If you need to keep image observations and other state files of different exploration runs, you may need to pass a different output_dir to each toolkit instance, and make EnvState accept that custom output_dir.

rpent/tools/toolkit.py
```diff
@@ -308,6 +308,18 @@ def raise_if_cancelled(self) -> None:
...
+        """Called when a fresh agent session takes over. Default: no-op."""
+
+    def solved(self) -> bool:
```
jx-qiu, line 314: One can query if a task is solved via EnvState APIs, so there should be no need for a dedicated API.

To query if a task is solved, use `toolkit.state.get().terminated`. This is under-documented and I will fix that.

rpent/tools/toolkit.py
Elessar123: Currently, each exploration session reconstructs the planner, and each session allows multiple attempts (for example, five) through environment resets, while reusing the same Toolkit across all sessions.

I am considering changing the lifecycle so that the Planner and Toolkit are reconstructed together at the beginning of each session. Attempts within the same session would still reuse that session's Toolkit.

With this design, `Toolkit.begin_session()` would no longer be necessary. I would move the persistent exploration state outside the Toolkit. Each session could use a separate state output directory, with a run-level component tracking and aggregating those session artifacts.

So the proposed lifecycle would be:
- one fresh Planner and Toolkit per exploration session;
- multiple reset-based attempts within that session;
- persistent run-level state/artifact management across all sessions.
Does this match the lifecycle you had in mind, or would you prefer reconstructing the Toolkit after every individual attempt as well?

rpent/tools/toolkit.py
Elessar123: Thanks! If you need any help, just call me.

rpent/tools/toolkit.py
jx-qiu: Your plan looks great!
> Does this match the lifecycle you had in mind, or would you prefer reconstructing the Toolkit after every individual attempt as well?

That depends on if you would like the on-disk stored images to be organized by session or by attempt. Both are OK from my perspective, and you can choose either one.

Elessar123: I wrote a version for session-level reconstruction; just check whether it is OK

docs/source-en/rst_source/usage/libero.rst
```diff
@@ -100,6 +100,56 @@ Minimal command
 
 To switch planners, see :doc:`configure_planner`.
 
+Exploration and local-memory evaluation
+---------------------------------------
+
```
qurakchin, line 105: add a clear description, such as:
we support 2 modes: exploration / evaluation,
1. what is exploration xxxxx
2. what is evaluation xxxxx
3. the evaluation is based on exploration result xxx

qurakchin: the success rate of harnessvla is reproduced on evaluation mod

robots/libero/__init__.py
```diff
@@ -38,20 +38,27 @@ def get_env_spec() -> EnvSpec:
...
         dashboard=LIBERO_DASHBOARD_SPEC,
+        finalize_run=_finalize_run,
     )
 
 
 def get_toolkit(
     *,
     primitives_kwargs: dict[str, Any],
     dashboard_events: DashboardEventSink,
+    explore: bool = False,
```
qurakchin: use attr name mode: str = 'evaluation'

robots/libero/__init__.py
```diff
@@ -72,6 +79,19 @@ def _add_cli_args(parser: argparse.ArgumentParser, use_dashboard: bool) -> None:
                         help="e.g. libero_object_task, libero_spatial_swap")
     parser.add_argument("--task", type=int, default=None, required=required)
     parser.add_argument("--seed", type=int, default=0)
+    parser.add_argument("--memory-profile", choices=["hf", "local"], default="hf",
```
qurakchin: move --memory-profile / --memory-dir / --explore to main.py

robots/libero/prompts/layered_eval.py
```diff
@@ -0,0 +1,75 @@
+"""Layered-memory variant of the single-attempt LIBERO evaluation prompt."""
```
qurakchin: what's the difference between the default eval mode and layered eval mode?

robots/libero/__init__.py
```diff
@@ -101,11 +121,25 @@ def _parse_config(args: argparse.Namespace) -> RunConfig:
...
         "task": args.task,
         "seed": args.seed,
         "recipe_tag": recipe_tag,
+        "mode": "explore" if explore else "eval",
+        "memory_profile": "local" if explore else args.memory_profile,
+        "memory_dir": str(memory_dir),
+        "reference_tag": f"{args.suite.replace('libero_', '')}_t{args.task}_s0",
+        # Per-cell inbox: parallel explore runs must not append to a shared file.
+        "memory_inbox": str(memory_dir / "_inbox" / recipe_tag),
+        "session_no": 1,
```
qurakchin: what's session_no? is this session_number?

robots/libero/__init__.py
```diff
@@ -101,11 +121,25 @@ def _parse_config(args: argparse.Namespace) -> RunConfig:
         raise ValueError("--task is required")
 
     recipe_tag = f"{args.suite.replace('libero_', '')}_t{args.task}_s{args.seed}"
+    explore = bool(getattr(args, "explore", False))
+    memory_dir = (
+        Path(args.memory_dir).expanduser().resolve()
+        if args.memory_dir
+        else get_memory_dir("libero")
+    )
```
qurakchin, line 132: are these situation available?
1. explore: true and memory-profile: hf
2. explore: true and explore-sessions <= 0
2. explore: false and memory-profile: hf and memory-dir: not None

robots/libero/prompt_bundle.py
```diff
@@ -2,13 +2,21 @@
...
 from robots.libero.prompts import system as system_parts
 from robots.libero.prompts import user as user_parts
 from rpent.context.prompt_utils import Numbered, PromptNode
 
 
-def system_prompt() -> PromptNode:
-    """Assemble the LIBERO system prompt tree."""
+def system_prompt(variables: Mapping[str, object] | None = None) -> PromptNode:
+    """Assemble the LIBERO system prompt for the selected run mode."""
+    if str((variables or {}).get("mode", "eval")) == "explore":
```
qurakchin: is the str().get() correct?

rpent/cli/main.py
```diff
@@ -289,7 +372,17 @@ def main() -> int:
     if agent_error:
```
qurakchin: delete this logger as the line 343 printed

robots/libero/prompts/explore.py
Elessar123: Agreed. I removed the additional recipe-name helper and kept the existing `write_recipe` interface. `write_recipe_from_states()` now accepts an explicit output directory, which lets a session-scoped Toolkit read its own `EnvState` while publishing the successful recipe at the TaskRun output root.

rpent/cli/main.py
Elessar123: Agreed. Dashboard and exploration can now be used together. But I find some bugs in the Dashborad, maybe I will fix it in the next PR

rpent/planner/claude_code.py
Elessar123: Yes, the two blocks performed the same check and only used different key spellings.I removed the duplicate `isError` assignment and kept only `is_error`.

robots/libero/prompts/layered_eval.py
Elessar123: Both are single-attempt evaluation modes. The default `hf` profile syncs and uses the prebuilt Hugging Face memory, while `local` reads the exploration-generated corpus from `--memory-dir`. I also renamed `layered_eval.py` to `local_eval.py`; “layered” now only describes the global/suite/task memory structure.

README.md
Elessar123: I also added fail-fast validation for this case. For a non-exploration run with `--memory-profile local`, RPent now requires `<memory-dir>/MEMORY.md` and raises a clear error if the local corpus is missing, suggesting either running exploration first or using `--memory-profile hf`.

docs/source-en/rst_source/usage/libero.rst
Elessar123: Thanks! I have added a clear two-mode description to the LIBERO documentation.

robots/libero/__init__.py
Elessar123: Agreed. I replaced `explore: bool = False` with `mode: str = "evaluation"` and pass `"exploration"` explicitly for exploration runs.

robots/libero/__init__.py
Elessar123: Thanks! I have moved them.

robots/libero/__init__.py
Elessar123: These cases are now handled explicitly:

`evaluation` + unspecified profile → use `hf`

`evaluation` + `hf` → valid

`evaluation` + `local` → valid, and require `MEMORY.md`

`exploration` + unspecified profile → use `local`

`exploration` + `local` → valid

`exploration` + `hf` → configuration error

`exploration` + `explore-sessions <= 0` → configuration error

`hf` + `memory-dir` → configuration error

robots/libero/__init__.py
Elessar123: Yes, it is the 1-based session number. I renamed `session_no` to `session_number` for clarity.

robots/libero/prompt_bundle.py
Elessar123: The original expression called `.get()` on the variables mapping and then converted the result with `str()`, but the parentheses made it easy to misread. The conversion was unnecessary, so I removed `str()` and now compare the mapping value directly.

rpent/cli/main.py
Elessar123: Agreed. I extracted the continuation-session setup into `_start_continuation_session()`.

rpent/cli/dashboard.py
```diff
@@ -152,59 +156,114 @@ def _run_dashboard_task(
...
+                        / f"session_{session_number:03d}"
+                    )
+                    state.begin_planner_session(
+                        video_path=state_output_dir / "episode.mp4",
+                    )
+                toolkit = get_toolkit(
+                    args.env_name,
+                    primitives_kwargs=primitives_kwargs,
+                    dashboard_events=state,
+                    mode=("exploration" if task_args.explore else "evaluation"),
```
qurakchin: only the get_toolkit function in libero support exploration now. so add an 'if' to filter other envs

Elessar123: Thanks! I have fixed it.

robots/libero/__init__.py
Elessar123: I have updated them

rpent/cli/main.py
Elessar123: OK

qurakchin (approved):

QuanluZhang (approved):

PR #80, opened 2026-08-10
refactor: add toolkit lifecycle hooks

Sonorous281:
## Summary

- add pre-execution and post-result hooks to Toolkit
- add a run finalization hook for environment-specific result handling
- keep existing toolkit and LIBERO behavior unchanged through default implementations

### DIFF
--- a/rpent/cli/main.py
+++ b/rpent/cli/main.py
@@ -1,3 +1,4 @@
+# Copyright 2026 RPent Contributors
 """Physical agent main CLI entrypoint."""
 # `rpent/cli/`
 #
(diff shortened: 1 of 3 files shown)


rpent/cli/main.py
```diff
@@ -266,6 +273,22 @@ def main() -> int:
         for d in daemons:
             d.stop()
 
+    try:
+        finalization = toolkit.finalize_run(
+            planner_result=planner_result,
+            planner_error=agent_error,
+        )
```
wilburx813: not sure is finalize_run() only meant to correct the result after the planner stops, or should it prevent the planner from finishing too early? if it's latter, finalize_run() is too late since env is already stopped, if it's former, I think it's fine

wilburx813 (approved):

PR #88, opened 2026-08-14
refactor: add env / vla base class

qurakchin:


rpent/tools/env_facade_base.py
```diff
@@ -0,0 +1,96 @@
...
+        self._rpc["env.step"] = self.step
+        self._rpc["env.chunk_step"] = self.chunk_step
+        self._readonly_methods.add("env.get_env_meta")
+
+    def _dispatch(self, method: str, args: tuple, kwargs: dict) -> Any:
+        handler = self._rpc.get(method)
+        if handler is None:
+            raise ValueError(f"unknown RPC method: {method!r}")
+        if method in self._readonly_methods:
+            with self._rwlock.read():
```
Sonorous281: Should this be _dispatch_lock? _rwlock is never initialized, so this will fail on the RPC call.

rpent/tools/env_facade_base.py
```diff
@@ -0,0 +1,96 @@
...
+
+    EGL single-thread:
+        Subclasses that must keep EGL single-threaded must override ``serve``
+        and dispatch everything to a dedicated render thread, so the MuJoCo EGL
+        context stays on the same thread. See robocasa's override for reference.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._dispatch_lock = RWLock()
```
Sonorous281, line 40: Do we need another RWLock here? RpcFacade.serve() already serializes all _dispatch() calls with a Lock, so this currently doesn't provide concurrent reads.

rpent/tools/env_client_base.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, client, *, expected_meta: dict):
+        self._client = client
+        server_meta = self._client.call("env.get_env_meta",
+                                        timeout_s=self._TIMEOUT_S["default"])
+        assert server_meta == expected_meta, (
+            f"env_meta mismatch: expected={expected_meta!r} actual={server_meta!r}. "
+            "The env_server was launched with different args than this client "
+            "expects — kill the stale env_server and relaunch."
+        )
+        self.last_obs = self.reset()
```
Sonorous281: LIBERO reset currently returns (obs, info), so self.last_obs = self.reset() would store the whole tuple. Should the base class handle the existing reset contract?

jx-qiu (approved):
It would be nice to see how these base classes and extracted env runtime functions are used, but let's prioritize merging and sort things out later.

jx-qiu:
Just one thing, what about renaming facade->server?

Sonorous281 (approved):

rpent/tools/env_facade_base.py
qurakchin: now the libero uses the old  RpcFacade. it need another pr to refactor it.

rpent/tools/env_facade_base.py
qurakchin: fixed

rpent/tools/env_client_base.py
qurakchin: fixed. the assign is only in the reset function now.

PR #103, opened 2026-08-24
fix(robotwin): dashboard optimization

Sonorous281:
Improve RoboTwin Dashboard support.

Fixes: `not solved` label after a successful task, missing action timeline, missing draggable video progress bar, and static images instead of the full dynamic video. Reasoning display and tool-call grouping are general planner behavior outside this PR's scope.

Aligns RoboTwin with the existing LIBERO/RoboCasa Dashboard contract.


robots/robotwin/primitives.py
```diff
@@ -144,7 +165,12 @@ def apply_qpos_updates(
                 action[offset : offset + 6] = update["arm_qpos"]
             if "gripper" in update:
                 action[offset + 6] = update["gripper"]
-            _, _, _, _, info = self.env.step(action, action_type="qpos")
+            step_result = self.env.step(action, action_type="qpos")
+            info = step_result[4]
```
QuanluZhang: suggest writing like this:
`obs, _, _, _, info = self.env.step(action, action_type="qpos")`

robots/robotwin/primitives.py
```diff
@@ -184,7 +210,19 @@ def lingbot_act(
             native_prompt = observation["task_language"]
             actions = self.model.infer(observation)[: MODEL_SPEC.use_length]
             self._check_cancelled()
-            _, _, _, _, info = self.env.chunk_step(actions, action_type="ee")
+            result = self.env.chunk_step(
```
QuanluZhang: same here

robots/robotwin/toolkit.py
```diff
@@ -194,8 +198,23 @@ def get_env_state(
...
+            if frames:
+                self._state.save(
+                    f"action_{command['action']}.mp4",
+                    frames,
+                    step=record.step_idx,
+                    fps=20,
+                )
         return tools.view_env_state(record.step_idx, state=self._state)
 
+    def close(self) -> None:
```
QuanluZhang, line 212: `close` is related to frame recording, right? but frame recording related functions are placed in primitives.py while `close` function is placed in toolkit.py. A little strange.

QuanluZhang (approved):

robots/robotwin/tools.py
```diff
@@ -330,14 +341,13 @@ def dump_observation(
...
-    # create training termination signals.
+    eval_success = status.get("eval_success") is True
     with env_state.record_step(
         state=state,
-        terminated=False,
+        terminated=eval_success,
         truncated=False,
         command=(log or {}).get("command"),
-        result=(log or {}).get("result"),
+        result=_to_json_safe((log or {}).get("result")),
```
wilburx813: maybe put _to_json_safe to common dashboard code? other env may encounter same problem

robots/robotwin/env_server.py
```diff
@@ -162,15 +160,25 @@ def chunk_step(
...
         if len(observation_list) != 1 or len(info_list) != 1:
             raise RuntimeError(
                 "RoboTwin chunk_step must return one environment, got "
                 f"{len(observation_list)} obs / {len(info_list)} info"
             )
+        observation = observation_list[0]
+        if return_all_frames:
+            observation = {
+                "frames": observation["frames"],
+                "final": self._strip_single_env_observation(observation["final"]),
```
wilburx813, line 176: This payload seems cleaner than list[Obs]. Could we make it the common return_all_frames contract and update other env accordingly?

robots/robotwin/toolkit.py
Sonorous281: close() is the toolkit lifecycle hook, while primitives only owns the frame buffer. libero and robocasa use the same split as well. I’ll keep the current structure to stay aligned with them.

robots/robotwin/env_server.py
Sonorous281: Agree this could be a cleaner common contract. LIBERO still uses list[Obs], but changing that is outside this PR’s scope, so we’d prefer to handle it separately.

wilburx813 (approved):

PR #109, opened 2026-08-26
refactor: refactor rpc / components path

qurakchin:


rpent/robots/components/pi05_vla_server.py
Sonorous281, line 45: This still looks LIBERO-specific (pi05_libero, 7 action_dim). Should it stay under robots/libero until it is actually shared across robots?

jx-qiu (approved):
The "components" naming in rpent/robots/*components* feels a bit too broad. Otherwise the PR looks good.

rpent/robots/components/pi05_vla_client.py
Sonorous281: Could we also update the remaining old server paths? The docstring and libero docs still reference robots/libero/vla_server.py.

Sonorous281 (approved):

PR #116, opened 2026-08-27
fix(robocasa): preserve rendered frame orientation

ucfyao:
## Summary

- stop rotating already top-down `render_camera()` frames a second time when recording
- add focused regression coverage for the recorded-frame orientation

## Why

`RoboCasaEnvClient.render_camera()` already converts MuJoCo output to top-down orientation. `RoboCasaPrimitives.record_frame()` then applies `[::-1, ::-1]`, rotating the frame by 180 degrees before it is written to the episode video.

The change preserves the orientation returned by `render_camera()` while retaining a contiguous `uint8` array for video encoding.

This was also exercised in two GPU RoboCasa runs (`ReheatMeatOnStove`, seed 1001) using RLDX and Xiaomi-MiBot backends. Both videos decoded successfully and an extracted frame was visually verified upright.

## Validation

```text
pytest -q tests/test_robocasa_camera_compat.py
1 passed

git diff --check
passed
```

### DIFF
--- a/tests/test_robocasa_camera_compat.py
+++ b/tests/test_robocasa_camera_compat.py
@@ -1,0 +1,22 @@
+import numpy as np
+
+from robots.robocasa.primitives import RoboCasaPrimitives
+
+
+class _RenderedEnv:
+    def __init__(self, frame: np.ndarray) -> None:
+        self.frame = frame
+
+    def render_camera(self, **kwargs) -> np.ndarray:
+        return self.frame
+
+
+def test_record_frame_keeps_render_camera_orientation() -> None:
+    frame = np.arange(18, dtype=np.uint8).reshape(2, 3, 3)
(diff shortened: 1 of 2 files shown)


qurakchin:
Thanks for the contribution.

tests/test_robocasa_camera_compat.py
```diff
@@ -0,0 +1,22 @@
...
+
+class _RenderedEnv:
+    def __init__(self, frame: np.ndarray) -> None:
+        self.frame = frame
+
+    def render_camera(self, **kwargs) -> np.ndarray:
+        return self.frame
+
+
+def test_record_frame_keeps_render_camera_orientation() -> None:
```
qurakchin, line 14: Strange test. I think this test is not need.

zoeyuchao:
This issue is fixed via #121, so we close this pr. Thanks for your contribution!

PR #117, opened 2026-08-28
fix(robotwin): work around place_fan reward construction bug

huanming99:
## Summary

  Add a temporary compatibility workaround for a reward-construction bug in the
  pinned RoboTwin `place_fan` environment.

  The pinned `place_fan` implementation still constructs `Reward` with the legacy
  `subtasks` and `transition_rewards` keyword arguments, while the corresponding
  `Reward` implementation now expects callers to use `Reward.build_top()`.

  This causes the RoboTwin EnvServer to exit during `place_fan` initialization.
  From RPent, the failure appears as an EnvServer ready timeout.

  ## Fix

  Install a narrowly scoped compatibility adapter at the RPent EnvServer boundary
  before constructing the pinned RoboTwin environment.

  The adapter:

  - applies only to `place_fan`;
  - converts the legacy arguments into the equivalent serial
    `Reward.build_top()` configuration;
  - is idempotent;
  - does not modify RoboTwin source code or dependency pins;
  - can be removed after the pinned RoboTwin dependency includes an upstream fix.

  ## Validation

  - Local focused regression test passed.
  - Integration check against the actual pinned RoboTwin `Reward` implementation
    passed.
  - `place_fan` Env-only initialization passed on the latest RPent `main`:
    - task: `place_fan`
    - task config: `demo_randomized`
    - seed: `100002`
    - EnvServer reached ready in 86.1 seconds
    - `healthz` succeeded
    - `env.get_env_meta` returned the expected task, config, seed, action layouts,
      and step limit
  - No reset, step, VLA inference, planner, MCP server, or episode was run.
  - The validation was limited to `place_fan` environment construction, server
    readiness, and runtime metadata.

### DIFF
--- a/robots/robotwin/env_server.py
+++ b/robots/robotwin/env_server.py
@@ -38,6 +38,7 @@ from omegaconf import OmegaConf  # noqa: E402
 from robotwin.assets import validate_root  # noqa: E402
 from robotwin.config import load_task_config  # noqa: E402
 
+from robots.robotwin.reward_compat import install_native_reward_compat  # noqa: E402
 from robots.robotwin.rlinf_env import RoboTwinAgentEnv  # noqa: E402
 from robots.robotwin.robot_spec import RoboTwinActionType  # noqa: E402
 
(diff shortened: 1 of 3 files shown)


Sonorous281:
Since the root cause is in RoboTwin's RLinf_support runtime itself, could we also open an upstream RoboTwin issue to track the place_fan Reward construction mismatch? Then this compatibility workaround in rpent can remain temporary.

Sonorous281 (approved):

QuanluZhang (approved):

PR #118, opened 2026-08-28
fix: harden tool result and failure handling

wilburx813:
## Summary

- keep truncated tool results within the configured UTF-8 byte limit
- roll back failed environment steps, remove their artifacts, and propagate the original exception
- cancel active toolkit operations after Claude or Codex planner timeouts

## Validation

- pre-commit run --all-files
- focused regression cases pass in the stacked test worktree

## Follow-up

Regression coverage will be included in the test baseline PR.


rpent/session/base.py
```diff
@@ -329,13 +329,24 @@ def record_step(
...
+                        cleanup_error,
+                    )
             if self._steps and self._steps[-1] is record:
                 self._steps.pop()
             logger.warning(
                 "step %d is discarded due to exception during record: %s",
                 record.step_idx,
                 e,
             )
+            raise
```
QuanluZhang: how about raising with a concrete error message?

QuanluZhang (approved):

qurakchin (approved):

PR #124, opened 2026-08-28
feat(planner): add configurable SDK agent runtimes

ye1457:
## Summary

- add `AgentRuntimeConfig` so SDK-backed planners can share an explicit repository root and isolated working directory
- forward custom endpoints and runtime context to both Claude Agent SDK and Codex SDK backends, including `claude-opus-4-8` model selection
- harden Codex Responses-compatible endpoint configuration with URL validation, selectable sandboxing, and secure `CODEX_API_KEY_FILE` loading
- document the shared extension point for future SDK-backed agent planners in English and Chinese

## Motivation

Experiment launchers need to provision isolated worktrees without coupling the launcher to one agent SDK. This change keeps backend-specific controls on each planner while giving Claude Code, Codex, and future agent backends a small common filesystem-context contract. It is independently mergeable preparatory work for #21.

## Security and compatibility

- no credentials, provider-specific secrets, or local paths are included
- Codex key files must be regular non-symlink files with exact `0600` permissions; key values are passed through the child environment rather than config overrides or logs
- custom base URLs reject embedded credentials, query strings, fragments, and `/responses` operation URLs
- the existing API planner and default SDK planner behavior remain unchanged when the new options are not supplied

## Verification

- `pytest -q`: 12 passed
- `ruff check .`: passed
- `ruff format --check .`: passed
- English Sphinx build with `-W`: passed
- Chinese Sphinx build completed with only the pre-existing upstream warning in `configure_primitives.rst:22`, a file unchanged by this PR

The Claude path is covered by SDK option and forwarding tests. No live Claude provider credential was used or included.

Issue #21, opened 2026-07-21
Questions about RoboCasa365 support, benchmark hardware/time, and LLM cost


coolmaoguo:
Hi RPent team,

Thanks for the great work!

I have three questions about RoboCasa365 and the reported benchmark:

1. The README mentions `scripts/setup_robocasa.sh`, `scripts/run_robocasa.sh`, and `docs/SETUP_ROBOCASA.zh.md`, but I cannot find them in the current `main` branch. Is the RoboCasa365 integration already open-sourced? If not, when is it expected to be released?

2. Could you share the hardware setup and evaluation time for the reported benchmark results? For example, GPU type/count, parallelization setting, time per task/seed, or total evaluation time.

3. Could you provide an approximate LLM API cost for running the benchmark? 
Thanks!

jx-qiu:
Hi coolmaogo,
1) Robocasa support is planned and will be available in the next few weeks.
2) You will need ~10-12 GiB of free GPU memory to run RPent on LIBERO Pro. Parallelization settings don't affect the results, so choose it according to available GPU memory and MaaS API concurrency. A single LIBERO run typically takes 3-10 minutes to complete, depending on the LLM used.
3) Using Claude Sonnet, each run costs about ~$0.5-2.

liberow:
Hi RPent team,
I have one more question after reading the Harness VLA paper:

The paper mentions using **RLDX-1** (the frozen VLA checkpoint from arXiv:2605.03269) as the vla_act backend for RoboCasa365. I noticed in the README Feature Matrix that RLDX-1 is listed without a ✅ checkmark. Is the RLDX-1 model checkpoint open-sourced? If so, where can we find the weights? If not, are there plans to release it, or will you support other open VLA backends (e.g., π₀.₅) for RoboCasa365 evaluation?

Thanks!

jx-qiu:
@qurakchin do you know when RLDX-1 will be released?

felixzheng02:
Hi RPent team,

Thanks for the great work.

Following up on the RoboCasa365 timeline: you mentioned a few weeks back in July, and I wanted to check whether that's still roughly the shape of it.

Thanks!

jx-qiu:
Hi Felix,

RoboCasa integration is almost ready. Code (see https://github.com/RLinf/RPent/pull/81) is already under review and is expected to be merged into mainline by next week.

PR #125, opened 2026-08-28
fix(robocasa): harden environment and VLA contracts

ye1457:
## Summary

- add explicit, versioned runtime handshakes for the RoboCasa environment and RLDX VLA RPC services
- validate environment actions, VLA observations/actions, modality metadata, session IDs, runtime IDs, and model fingerprints before execution
- latch task success after every simulator action and stop action chunks, analytic primitives, and VLA rollout chunks immediately after success
- enforce perception-isolated RPC/tool surfaces without exposing task progress or success predicates
- expose camera metadata and bounded pixel-to-world back-projection tools, with calibration cache invalidation after robot/base motion

## Why

The initial RoboCasa integration is functional, but long benchmark runs need fail-fast contracts around independently spawned environment and VLA services. Without these checks, stale sessions, malformed action batches, post-success motion, or privileged simulator state can silently invalidate a cell.

This PR is intentionally limited to the runtime boundary. Checkpoint attestation, packaged dependency patches, immutable memory resources, the frozen benchmark protocol, and the resumable reproduction CLI are kept in follow-up PRs. It is independent of #124 and is part of the RoboCasa release work tracked by #21.

## Compatibility

- ordinary RoboCasa runs retain the existing CLI and server startup flow
- formal perception isolation is opt-in through the runtime environment
- no planner credentials, model weights, memory artifacts, local paths, or private launch scripts are included

## Verification

- `pytest -q`: 19 passed
- focused runtime contract suite: 18 passed
- `ruff check .`: passed
- `ruff format --check .`: passed

### DIFF
--- a/robots/robocasa/env_client.py
+++ b/robots/robocasa/env_client.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from typing import TYPE_CHECKING
 
 import numpy as np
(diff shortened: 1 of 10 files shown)


ye1457:
Superseded by the direct-to-main minimal PR #130. The replacement keeps construction valid while `env.sim` is absent and validates `mobilebase0_navview` only after the first reset. This branch remains available for history, but it is no longer in the merge chain.

PR #126, opened 2026-08-28
feat(robocasa): add reproducible resources and model identity

ye1457:
> **Stacked PR:** depends on #125. The incremental commit is `72d2b89`; after #125 merges, this PR diff will automatically reduce to the resource/identity layer.

## Summary

- publish an immutable checksum authority for the audited `RLDX-1-FT-RC365` model and VLM metadata
- verify checkpoint contents once, emit a private reusable attestation, and bind VLA startup/runtime handshakes to the verified fingerprint
- package all `robots.*` backends plus the RoboCasa checkpoint manifest and delivery patch in built wheels
- pin `RLinf/robosuite` to the actually tested commit `85abee228d1c43ab1939bce33028099945d453b4`
- ship and test the exact `navview` XML patch required by perception-isolated RoboCasa runs
- exercise RoboCasa Hugging Face resource sync through the existing `ensure_resources` path, with explicit local-memory mode instead of silent fallback

## Reproduction boundary

This repository still does **not** contain model weights, RoboCasa assets, planner credentials, or memory payloads. Reproducers provision those inputs themselves. The future memory dataset is selected through the existing robot resource repository contract and will be downloaded at runtime; an offline local path requires explicit local mode.

The checkpoint attestation is generated locally after hashing user-provisioned files. It must be an owner-controlled regular file with mode `0600`, and the VLA process validates both the authority and file identity before accepting it.

The benchmark protocol and runner are deliberately excluded from this PR and remain follow-up layers for #21.

## Verification

- `pytest -q`: 31 passed
- checkpoint, navview, packaging, and runtime subset: 30 passed
- two wheels were built and inspected during tests
- `ruff check .`: passed
- `ruff format --check .`: passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,7 @@ robocasa = [
     "rpent[rlinf]",
     "rlinf-robocasa365",
     "rlinf-rldx>=1.0.1.post10",
-    "robosuite @ git+https://github.com/RLinf/robosuite.git@rlinf",
+    "robosuite @ git+https://github.com/RLinf/robosuite.git@85abee228d1c43ab1939bce33028099945d453b4",
 ]
 full = [
     "rpent[rlinf,openpi,libero-pro,sam3,robocasa]",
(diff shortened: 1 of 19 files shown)


ye1457:
Superseded by #130 for the basic integration. The replacement reuses the existing `ensure_resources()` flow and only selects the current task seed-0 pair; model identity and the larger reproducibility resource framework are deferred to later work. This branch is retained for that future design.

PR #131, opened 2026-08-31
Fix: sam3 rpc

6UOOON9:
## Summary

This PR fixes the bug:

SAM3 RPC readiness fails because `Sam3Facade._dispatch()` bypasses the base `RpcFacade` handling of lifecycle methods such as `healthz`.

## Problem
Sam3Facade._dispatch() only handled segment and raised directly for every other method. After RPC lifecycle handling was lifted into RpcFacade, this bypassed the base implementation. As a result, healthz returned:
```
unknown RPC method: 'healthz'
```
Dashboard readiness checks therefore never succeeded, causing the session to wait for 300 seconds and enter a fatal state. Since env_server is task-scoped and starts after shared services become ready, it was never spawned.

## Fix
- Change the SAM3 dispatch fallback from raising an error to calling `super()._dispatch()`.
- This restores base lifecycle methods such as `healthz` and `shutdown`.
- Existing `segment` behavior is unchanged.

## Testing
SAM3 server starts successfully.

### DIFF
--- a/rpent/robots/components/sam3_server.py
+++ b/rpent/robots/components/sam3_server.py
@@ -304,7 +304,7 @@ class Sam3Facade(RpcFacade):
     def _dispatch(self, method: str, args: tuple, kwargs: dict) -> Any:
         if method == "segment":
             return self.segment(*args, **kwargs)
-        raise ValueError(f"unknown RPC method: {method!r}")
+        return super()._dispatch(method, args, kwargs)
 
     def segment(
         self,


wilburx813 (commented):
Thanks for your contribution! The fix for Bug 2 looks great. 
For Bug 1, I think it would be better fixed upstream in https://github.com/RLinf/openpi/tree/rpent, rather than monkey-patch in RPent. Would you be willing to open a pr there? If you don’t have the bandwidth, please let us know and we can take it over.

6UOOON9:
Thanks so much for the review! I'm glad the fix for Bug 2 looks good.

I also agree that fixing Bug 1 upstream in RLinf/openpi would be the better approach, since it addresses the dtype mismatch at its source rather than monkey-patching it in RPent. Unfortunately, I'm not very familiar with the OpenPI codebase yet, so I'd appreciate it if you could take it over. Please feel free to adapt or reuse the patch there if helpful.

Thank you again for maintaining such an excellent project!

wilburx813:
> Thanks so much for the review! I'm glad the fix for Bug 2 looks good.
> 
> I also agree that fixing Bug 1 upstream in RLinf/openpi would be the better approach, since it addresses the dtype mismatch at its source rather than monkey-patching it in RPent. Unfortunately, I'm not very familiar with the OpenPI codebase yet, so I'd appreciate it if you could take it over. Please feel free to adapt or reuse the patch there if helpful.
> 
> Thank you again for maintaining such an excellent project!

Thanks for your understading, we’ll take over the bug1 fix in upstream. could you please keep this pr focused on bug2? Thanks again for your contribution!

wilburx813 (approved):

qurakchin (approved):

qurakchin:
Thanks for your contribution!

PR #132, opened 2026-08-31
feat: support session-aware rpc client / server

qurakchin:
support session-aware rpc client / server
use session-aware rpc client / server in robocasa
add a mixin for main-thread serve mode


wilburx813 (approved):

wilburx813 (commented):
The _dispatch() overrides in sam3 and robotwin don’t accept session_id and will cause failure. but these overrides also seem unnecessary, maybe cleanup in this pr?

robots/robocasa/rldx_skill.py
```diff
@@ -375,9 +377,13 @@ def run(
     def reset_session(self):
```
Sonorous281, line 377: Could we also call reset_session when a new RoboCasa TaskRun starts? The dashboard reuses the shared VLA client and session across tasks, so otherwise policy state from the previous task may carry over.

docs/source-zh/rst_source/development/add_primitive.rst
```diff
@@ -149,6 +149,61 @@ primitives 方法，以及调用完成后的状态快照。区别仅在于方法
 如果模型会保存每个回合的内部状态，应提供 ``vla_reset`` RPC，并在任务之间
 调用它完成重置。这样，同一个服务进程就能安全地复用于多次连续运行。
 
+session-aware 的 VLA 后端（per-client 策略状态）
+------------------------------------------------
```
Sonorous281, line 153: This section feels a bit unnatural and mixes Chinese and English heavily. Could we make the wording more idiomatic and consistent?

robots/robocasa/rldx_skill.py
qurakchin: policy states between different envs are isolated by session_id. The session_id is registered differently from different envs.

docs/source-zh/rst_source/development/add_primitive.rst
qurakchin: fixed

tests/utils/rpc/_rpc_test_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+from rpent.utils.rpc.http_rpc import HttpRpcClient, HttpRpcServer
+from rpent.utils.rpc.socket_rpc import SocketRpcClient, SocketRpcServer
+
+# Keep urllib off the host's system proxy. ``urllib.request.urlopen`` consults
+# the macOS system proxy (e.g. a local Clash at 127.0.0.1:7892); routed
+# through it, calls to 127.0.0.1 get a 502 / empty body. ``no_proxy`` forces
+# a direct connection so the HTTP transport is deterministic in tests.
+os.environ["no_proxy"] = "*"
+os.environ["NO_PROXY"] = "*"
```
Sonorous281: Could we avoid setting NO_PROXY=* globally at module import time? This affects networking for the entire test process, and the Clash-specific comment also looks environment-specific. Could the loopback proxy bypass be scoped to these tests instead?

docs/source-en/rst_source/development/add_primitive.rst
```diff
@@ -166,6 +166,69 @@ If the model keeps per-episode state, expose a ``vla_reset`` RPC and
...
+  ``_on_session_drop`` — clean up that client's policy state when the session
+  ends (the client's ``session.close`` RPC or idle expiry). If you need an
+  explicit reset, expose an extra ``reset_session`` RPC (clears policy state
+  only, does not destroy the session). ``serve`` must pass ``session_sweep_s``
+  (> 0) so a background thread periodically reclaims expired sessions.
+
+- **Client side**: construct the ``RpcClient`` inside the model client with
+  ``enable_sessions=True``; it registers a session with the server on
+  connect. ``session_id`` is derived from the connection and injected into
+  the server-side handler by the facade — the client does **not** pass it,
```
Sonorous281, line 190: session_id is generated by RpcClient and sent in each transport request, so “derived from the connection” / “the client does not pass it” seems a bit misleading. Could we clarify that business code does not pass it explicitly?

robots/robocasa/rldx_skill.py
Sonorous281: Could you clarify where a new session_id is created for each Dashboard TaskRun? The VLA client is shared across tasks, so it looks like the same session is reused. If the first reset_memory=True is intended to fully reset all RLDX session state, is that equivalent to reset_session() here?

wilburx813 (approved):

docs/source-en/rst_source/development/add_primitive.rst
qurakchin: The client means the caller who use the RpcClient, so the session_id is passed by the RpcClient, not the caller client.

tests/utils/rpc/_rpc_test_helpers.py
qurakchin: NO_PROXY=* here are removed

Sonorous281 (approved):

PR #120, opened 2026-08-28
test: add offline contract suite and CI workflows

wilburx813:
## Summary

- add a minimal `test` extra to `pyproject.toml` with `pytest` and `pytest-timeout`, plus the source-checkout import path and a 60-second per-test timeout
- establish an offline contract-test suite for core toolkit and state behavior, all three robot integrations, API/Claude/Codex planners, Dashboard control and session/launcher state, layered memory and its maintenance CLI, and the complete CLI orchestration path
- add a Unit tests workflow that runs the suite independently on Python 3.10, 3.11, and 3.12 for pull requests and pushes to `main`
- add English and Chinese contribution guides, a concise pull-request template, and README links so contributors can run and report the same checks locally

## Remaining tasks

- [x] merge #118, update this branch to the latest `main`, confirm that no production-code changes remain, and rerun the final checks

## Validation

- [x] `pre-commit run --all-files`
- [x] Python 3.10 — 161 passed in a clean environment
- [x] Python 3.11 — 161 passed in a clean environment
- [x] Python 3.12 — 161 passed in a clean environment
- [x] 161 passed with adversarial repository, planner, credential, robot, and runtime environment variables
- [x] latest `main` on Python 3.12 — 171 passed
- [x] English documentation build with warnings treated as errors
- [x] Chinese documentation build with warnings treated as errors

PR #140, opened 2026-09-01
fix(robocasa): enforce live task language for VLA execution

ye1457:
## Summary

- always send RoboCasa live `task_language` to RLDX, even when historical memory supplies an atomic prompt or `use_prompt=true`
- remove `use_prompt` from planner-visible VLA schemas while retaining handler compatibility for historical recipes
- record `effective_prompt` / `prompt_overridden` and keep VLA continuity with bounded re-staging
- normalize null image geometry in the published RLDX-1 checkpoint to the release defaults before inference

## Scope

- RoboCasa runtime, tool schema, prompt, and focused tests only
- no planner-specific code, shared CLI/RPC/resource changes, memory dataset changes, evaluation manifest, or runner

## Conflict resolution

- rebased onto `upstream/main` at `1b5bc42`
- adopted the offline suite layout introduced by #120
- kept `tests/unit_tests/robots/robocasa/test_resource_contracts.py` identical to `main`
- moved only the seven #140-specific checks into `test_vla_protocol_contracts.py`; the removed legacy test path was not restored
- verified all four RoboCasa runtime file blobs are byte-for-byte identical to the previously GPU-tested `10b24e4` version

## Validation

- `pytest -q tests/unit_tests/robots/robocasa/test_vla_protocol_contracts.py`: 7 passed
- `pytest -q tests/unit_tests/robots/robocasa`: 14 passed
- `pytest -q tests/unit_tests`: 192 passed
- `pre-commit run --all-files`: passed locally
- tracked-file secret, private-endpoint, and server-path scan: clean
- default HF sync fetched the published 111-file RoboCasa corpus
- real RTX 4090 env/VLA RPC smoke: RGB-D/world map finite, navview follows the mobile base, and an RLDX action chunk executes without preprocessing errors

The host used for local tests exports `HTTP_PROXY` / `HTTPS_PROXY`, so localhost RPC tests were run with `NO_PROXY=127.0.0.1,localhost`. This is a host configuration requirement; this focused PR does not modify shared RPC behavior.

| Task | Seed 1 | Seed 2 |
| --- | --- | --- |
| `OpenDrawer` | success | success |
| `NavigateKitchen` | success | success |
| `PickPlaceCounterToCabinet` | success | success |

The representative Atomic gate passed 6/6 with Codex SDK, `gpt-5.5`, `xhigh`, and `max_turns=100`. Every table entry is a final clean run whose success came from the returned environment state, not `finish(status=...)`. `OpenDrawer` seed 1 passed as required. Infrastructure-only attempts that produced no valid result were excluded and rerun.

After the review update, `OpenDrawer` seed 1 was run again at the runtime version retained by this branch. The first arm call reached `cap` without contact, the planner continued with the same full task language, and the second call returned environment `success=true` with `joint_p=1.065`. This confirms the shortened 2-3-call continuity guidance without relying on the finish hook.

## Upstream RLDX follow-up

The processor-boundary fix is isolated and covered by four regression cases in [`ye1457/RLDX-1@48dc1b9`](https://github.com/ye1457/RLDX-1/commit/48dc1b9). The currently released `rlinf-rldx` package still needs the narrow RPent fallback; it can be removed after the upstream change is merged and released.

## Follow-up

After this PR merges, a separate PR based on the new `main` will add the immutable Target50 manifest, the dedicated RoboCasa README, and reproduction documentation. This PR does not claim completion of the 340-cell Target50 reproduction.

### DIFF
--- a/robots/robocasa/prompts/system.py
+++ b/robots/robocasa/prompts/system.py
@@ -117,13 +117,10 @@ VLA EXECUTION RULES (the single most important rule):
    command WIPES the VLA's frame history, forcing it to restart from scratch.
 3. Do NOT pass a small max_chunks (default 70 is fine). Do NOT pass
    settle_patience (default 999 disables settle detection).
-4. After each VLA call, inspect its result, current RGB-D, and task_progress. If
-   there is contact, a held object, fixture progress, or a rising task counter,
-   continue with the same full task language and do not insert a manual command.
-5. If there is confirmed no-contact and no visible semantic progress, end that
-   VLA attempt. Re-localize, make only a bounded free-space re-stage, then begin
-   a fresh VLA attempt; do not repeat forever from the same ineffective pose.
-6. The vla_desync flag in view_env_state tells you if the VLA history was
+4. If RLDX returns 'cap', call it again with the same full task language to
+   preserve history. Only re-stage after 2-3 consecutive calls show neither
+   contact nor task progress.
+5. The vla_desync flag in view_env_state tells you if the VLA history was
    invalidated. If True, the next VLA call will start fresh.
 """
 
--- a/robots/robocasa/prompts/system.py
+++ b/robots/robocasa/prompts/system.py
@@ -110,15 +110,20 @@ PERCEPTION TOOLS:
 
 VLA_RULES = """
 VLA EXECUTION RULES (the single most important rule):
-1. NEVER put a manual command (move_to / move_base / navigate_to / set_gripper /
+1. Every rldx_skill / rldx_arm call must pass the complete live task_language
+   verbatim. Never shorten, paraphrase, or replace it with an atomic sub-task.
+2. NEVER put a manual command (move_to / move_base / navigate_to / set_gripper /
    scripted_grasp) BETWEEN two VLA calls of the same sub-operation. Every non-VLA
    command WIPES the VLA's frame history, forcing it to restart from scratch.
-2. Do NOT pass a small max_chunks (default 70 is fine). Do NOT pass
+3. Do NOT pass a small max_chunks (default 70 is fine). Do NOT pass
    settle_patience (default 999 disables settle detection).
-3. If rldx_skill returns 'cap' (didn't finish in budget), simply CALL IT AGAIN
-   with the same prompt — continuity is preserved. Only re-stance if 2-3 consecutive
-   calls never touched the object (grasp_detected was never true).
-4. The vla_desync flag in view_env_state tells you if the VLA history was
+4. After each VLA call, inspect its result, current RGB-D, and task_progress. If
+   there is contact, a held object, fixture progress, or a rising task counter,
+   continue with the same full task language and do not insert a manual command.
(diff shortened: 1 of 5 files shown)


wilburx813 (approved):
overall looks good, the fix rldx feels a bit monkey-patch, I think it's better to be fixed in rlinf-rldx, but I think it's ok to merge if it's urgent

robots/robocasa/prompts/system.py
```diff
@@ -110,15 +110,20 @@
...
    scripted_grasp) BETWEEN two VLA calls of the same sub-operation. Every non-VLA
    command WIPES the VLA's frame history, forcing it to restart from scratch.
-2. Do NOT pass a small max_chunks (default 70 is fine). Do NOT pass
+3. Do NOT pass a small max_chunks (default 70 is fine). Do NOT pass
    settle_patience (default 999 disables settle detection).
-3. If rldx_skill returns 'cap' (didn't finish in budget), simply CALL IT AGAIN
-   with the same prompt — continuity is preserved. Only re-stance if 2-3 consecutive
-   calls never touched the object (grasp_detected was never true).
-4. The vla_desync flag in view_env_state tells you if the VLA history was
+4. After each VLA call, inspect its result, current RGB-D, and task_progress. If
```
qurakchin: Is this hack too obvious

qurakchin: compare to the previous 'Only re-stance if 2-3 consecutive calls never touched the object'

ye1457: Agreed, this became too prescriptive. I collapsed it back to the previous bounded continuity rule: continue with the full task language, and only re-stage after 2-3 consecutive calls show neither contact nor task progress. The live task-language guarantee remains enforced by the runtime.

ye1457:
@wilburx813 Agreed that the normalization belongs in RLDX. I prepared the processor-boundary fix against current upstream `main` at [ye1457/RLDX-1@48dc1b9](https://github.com/ye1457/RLDX-1/commit/48dc1b9), with four regression cases covering null, partially null, and custom geometry values.

I also attempted to open the cross-fork PR, but the RLWRLD repository currently rejects `CreatePullRequest` for this account despite read access; the tested branch is ready at [`fix/null-image-geometry`](https://github.com/ye1457/RLDX-1/tree/fix/null-image-geometry). Until that source fix is merged and released as `rlinf-rldx`, #140 keeps only the narrow post-load fallback required by the currently published checkpoint/package combination. It can be removed once the upstream release contains the fix.

qurakchin (approved):

ye1457:
Resolved the conflict with the latest `main` by rebasing onto `1b5bc42`. I adopted #120's `tests/unit_tests` layout, kept the new resource contract file unchanged, and moved only the seven #140-specific checks into `test_vla_protocol_contracts.py`; the legacy test path was not restored.

The four RoboCasa runtime file blobs are identical to the previously reviewed and GPU-tested version. Local validation is 192/192 unit tests plus pre-commit, and the GitHub Pre-commit and Python 3.10/3.11/3.12 unit-test jobs are all green.

PR #141, opened 2026-09-02
feat(robocasa): define and document the target50 evaluation suite

ye1457:
## Summary

- define the immutable RoboCasa Target50 manifest for 18 Atomic, 16 Composite-Seen, and 16 Composite-Unseen tasks (340 cells)
- freeze the evaluated seed and RLDX runtime contract, compatibility-sensitive dependencies, Robosuite commit, RLDX-1 revision, and task-only memory snapshot
- define the RoboCasa environment-authoritative `result.json` schema, register it through #148's generic finalization hook, and provide a fixed-denominator validator with task-weighted aggregation
- document the complete source-checkout reproduction flow in the standalone README and aligned English/Chinese usage guides
- publish the supplied formal Target50 Codex results for per-task comparison

This PR is rebased directly on `upstream/main@0e4b8d0`, including the merged live-task-language fix in #140, hostname-based proxy fix in #144, RoboTwin mirror fix in #145, session-aware RPC support in #132, generic result finalizer in #148, and unified cross-robot memory refactor in #142. It does not add a planner-specific RoboCasa path, sandbox, supervisor, runner, or global memory.

## Proxy behavior

With the merged #144, HTTP RPC endpoints whose hostname is exactly `127.0.0.1` or `localhost` use request-local direct access, whether RPent starts the worker or the user supplies the endpoint. Every other hostname and IP uses standard environment proxy resolution. Codex scopes the same two-host exception to its child environment for the local MCP connection. Users keep `HTTP_PROXY` and `HTTPS_PROXY` for remote Hugging Face and planner traffic; the default runtime requires no shell-wide `NO_PROXY` setup. A custom local hostname can still be added to the user's existing `NO_PROXY` / `no_proxy` configuration when direct access is desired.

## Frozen protocol

- protocol: `robocasa-harness-vla-v1`
- scene identity: ordinary `--seed`; spawned RoboCasa workers clear legacy `RLDX_RESET_SEED`
- Target50 RLDX execution: `max_chunks=40`, `settle_patience=999`, `n_action_steps=8`; ordinary RoboCasa retains `max_chunks=70`
- Robosuite: `97cfbde4b68d8ec43dad20cf4747297866a6ca2e`
- RLDX-1: `587e9ecdcc5e7184fcc17f58713908edff5af041`
- RPent memory: `551fc3157b3e56b40a3d3a3b4c7ff81721ebe89b`
- reference planner: Codex SDK, `gpt-5.5`, `xhigh`, `max_turns=100`
[...]

PR #144, opened 2026-09-02
fix(rpc): bypass proxies for localhost HTTP endpoints

ye1457:
## Summary

- Make `HttpRpcClient` bypass HTTP proxies automatically for exact `127.0.0.1` and `localhost` hostnames.
- Keep every other HTTP RPC hostname on standard environment proxy resolution.
- Keep direct routing request-local; RPent does not mutate the parent `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, or lowercase variants.
- Keep the Codex local MCP readiness probe and child connection local while preserving user proxy settings for remote planner/model traffic.

## Routing contract

| Endpoint hostname | HTTP behavior |
| --- | --- |
| `127.0.0.1` | Direct |
| `localhost` (case-insensitive) | Direct |
| Any other hostname or IP | Standard environment proxy resolution |

There is no public or internal `proxy_mode` switch. The same hostname rule applies to RPent-started workers and user-supplied HTTP RPC endpoints. Socket/WebSocket transports are unchanged.

## Rationale

Local HTTP RPC and MCP connections must not be sent through a remote proxy, while remote/custom endpoints, Hugging Face access, and planner/model requests must retain the users proxy configuration. Centralizing the hostname rule gives LIBERO, RoboTwin, and RoboCasa the same behavior without robot-specific call-site flags.

## Validation

- Exact-host direct tests for `127.0.0.1` and `localhost` under an unreachable proxy with no `NO_PROXY`.
- Recording-proxy tests for a normal remote hostname and `127.0.0.2`.
- Classification coverage for uppercase `LOCALHOST`, `127.0.0.2`, IPv6 loopback, and a remote hostname.
- Codex child-environment and local MCP readiness regression tests.
- `204 passed` across `tests/unit_tests`.
- `pre-commit run --all-files` passed.
- Secret and absolute-path scan passed.

### DIFF
--- a/rpent/planner/codex.py
+++ b/rpent/planner/codex.py
@@ -58,6 +58,23 @@ logger = get_logger("codex")
 
 PROVIDER_ID = "rpent_proxy"
 PROVIDER_ENV_KEY = "RPENT_CODEX_PROVIDER_KEY"
+_LOOPBACK_NO_PROXY_HOSTS = ("127.0.0.1", "localhost")
+
+
+def _codex_environment() -> dict[str, str]:
+    """Build the Codex child environment with direct access to its local MCP."""
+    env = {**os.environ}
+    entries: list[str] = []
+    for key in ("NO_PROXY", "no_proxy"):
+        entries.extend(item.strip() for item in env.get(key, "").split(","))
+    entries = list(
+        dict.fromkeys(item for item in (*entries, *_LOOPBACK_NO_PROXY_HOSTS) if item)
+    )
+    bypass = ",".join(entries)
+    env["NO_PROXY"] = bypass
+    env["no_proxy"] = bypass
+    return env
+
 
 # ---------------------------------------------------------------------------
 # Public backend
@@ -436,7 +453,7 @@ class CodexPlanner:
     # -- config builder ----------------------------------------------------
 
     def _build_config(self, mcp_url: str) -> Any:
-        env = {**os.environ}
+        env = _codex_environment()
         if self._api_key:
             env[PROVIDER_ENV_KEY] = self._api_key
         kwargs: dict[str, Any] = {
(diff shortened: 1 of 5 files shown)


ye1457:
Closing this PR in favor of the standard `NO_PROXY` / `no_proxy` configuration now documented in #141. We want to retain the HTTP client's normal environment-proxy semantics instead of forcing direct connections in shared RPC code. Users keep `HTTP_PROXY` / `HTTPS_PROXY` for remote traffic and exempt only their local RPent worker hosts.

ye1457:
Reopened with a revised ownership-based design after validating the proxy matrix.

The old hostname heuristic has been removed. `HttpRpcClient` now defaults to `proxy_mode="environment"`; only clients for RPent-spawned HTTP workers opt into request-local `direct`. User-supplied endpoints continue to honor their environment proxy configuration. Codex receives loopback exclusions only in its child-process environment so remote model traffic still uses the user's proxy, and no parent/global environment is mutated.

Validation: 201 unit tests passed, pre-commit passed, and controlled dead-proxy tests cover environment/direct/user-`NO_PROXY` behavior. This supersedes the earlier close rationale and allows #141 to remove its universal `NO_PROXY` setup block.

ye1457:
Additional live proxy-path validation passed: with `HTTP_PROXY` / `HTTPS_PROXY` enabled and both `NO_PROXY` / `no_proxy` removed from the parent process, a real Codex SDK turn reached the remote model and completed one local MCP `finish` call successfully (`finish_status=success`, `tool_calls=1`). The parent process remained unchanged (`NO_PROXY` still absent). No credential or private endpoint is present in this branch.

ye1457:
Updated in `12985ec` to match the requested contract:

- removed `proxy_mode` and all robot-specific call-site flags;
- only `127.0.0.1` and `localhost` are automatically routed through request-local direct access;
- every other hostname/IP uses the standard environment proxy configuration;
- parent proxy environment variables are never mutated;
- retained the separate Codex child-only bypass for those same two hostnames because that MCP connection does not use `HttpRpcClient`.

Validation: exact-host direct tests, recording-proxy tests for a remote hostname and `127.0.0.2`, classification coverage including IPv6 loopback, Codex MCP regression tests, all `204` unit tests, and pre-commit all pass.

qurakchin (approved):

Sonorous281 (approved):

PR #152, opened 2026-09-06
feat(planner): add shared LLM backend connectivity check

JamesHuang0319:
## Summary

Add a lightweight LLM backend connectivity check that can be run before starting the robot/runtime stack.

The same check is shared by:

- a new `rpent-check-llm` CLI
- the Dashboard launcher via a **Test connection** button

This lets users catch model, credential, endpoint, SDK, and provider errors without first starting memory sync, env servers, VLA/SAM3, or robot runtime components.

## What changed

- Added a shared, UI-agnostic `check_llm()` layer for:
  - `api`
  - `claude_code`
  - `codex`
- Added `rpent-check-llm` as an independent CLI entry point.
- Added `POST /api/llm/check` to the Dashboard server.
- Added **Test connection** to the Dashboard launcher.
- Reused the same launcher configuration for both connectivity checking and session startup.
- Extracted shared API/Codex configuration helpers instead of duplicating provider construction logic.
- Added structured status classification and secret redaction.
- Added independent diagnostic timeouts.
- Added small Dashboard UX polish:
  - loading spinner + elapsed time
  - stale result clearing when configuration changes
  - race protection for in-flight checks
  - stable zh/en button sizing
  - spacing / launcher positioning polish

No new dependencies were added.

## Failure classification

The check reports structured statuses such as:

- `ok`
- `missing_config`
- `unsupported_provider`
- `missing_api_key`
- `auth_failed`
- `network_error`
- `provider_error`
- `sdk_error`

The check itself is designed not to raise errors to the CLI/GUI layer.

## Runtime isolation

The connectivity check intentionally does **not** start or require:

- robot runtime
- `MemoryManager.sync()`
- env server
- VLA
- SAM3
- Toolkit / MCP runtime
- simulator

It only sends the smallest real request required to verify the selected backend.

## Codex SDK compatibility

Real server validation exposed a Codex SDK integration issue where `TurnHandle`
was treated as directly iterable. The probe now uses the supported
`TurnHandle.run()` path, with regression coverage added for the SDK object shape.

## Testing

- Full unit test suite: `305 passed`
- `ruff check --preview` and `ruff format --check`: passed
- Server planner tests: `89 passed`

Manual validation was also performed against the real Codex backend:

- CLI connectivity check returned `ok: true` with a non-empty reply.
- Dashboard connectivity check succeeded with a supported Codex model and reported latency.
- An unsupported model was surfaced as a provider error in the UI without crashing the Dashboard or returning HTTP 500.
- Connectivity checks completed without starting the robot runtime.

## Out of scope

Model preset coverage and planner/provider/model configuration are not changed in this PR and can be addressed separately.


rpent/planner/base.py
```diff
@@ -114,6 +117,54 @@ def solve(
 # ---------------------------------------------------------------------------
 
 
+def build_api_model(model: str | None, base_url: str | None = None) -> "Model":
```
qurakchin, line 120: Is this function used in other files?

rpent/planner/check.py
```diff
@@ -0,0 +1,648 @@
...
+# ---------------------------------------------------------------------------
+
+
+def _check_claude_code(request: LlmCheckRequest) -> LlmCheckResult:
+    """Probe the Claude Agent SDK with one tool-free, single-turn query."""
+    model = (request.model or "").strip() or "sonnet"
+    key_present = bool(os.environ.get(_CLAUDE_CODE_KEY_ENV))
+
+    def _result(status: str, **kwargs: Any) -> LlmCheckResult:
+        return LlmCheckResult(
```
qurakchin, line 606: Will this line do a real llm api call in claude code?

qurakchin (commented):
add command description in docs/source-zh/rst_source/quickstart.rst;

rpent/planner/base.py
JamesHuang0319: Yes. It is used by both the normal build_planner() path and _check_api() in the connectivity check. It was extracted so both paths share the same provider/model construction logic

rpent/planner/check.py
JamesHuang0319: Not this line itself. _result only formats the result. The real Claude call happens in _run_claude_probe, which sends one minimal one-turn request through the Claude Agent SDK, with no tools, MCP, or robot runtime.

qurakchin:
1. The --base-url in main.py only affects the api. Change the help text in args and Raise an error when using claude code or codex.
2. fix the doc. --base-url is only for api.

rpent/planner/check.py
```diff
@@ -0,0 +1,946 @@
...
+            ok=status == STATUS_OK,
+            status=status,
+            planner="claude_code",
+            model=model,
+            credential_env=_CLAUDE_CODE_KEY_ENV,
+            credential_present=key_present,
+            **kwargs,
```
qurakchin, line 615: lack of base url. codex tests the base url.

qurakchin:
Running rpent-check-llm --planner codex with no api key will wait 90s then raise network_error. please raise fast if no api key
image

JamesHuang0319:
@qurakchin 
--base-url： Fixed. --base-url is now API-planner-only in both CLIs and the docs. Passing it with claude_code or codex raises an error instead of being silently ignored. The help text and English/Chinese planner docs were updated accordingly.  
inline lack of base url： Fixed. The claude_code check result now reports base_url and base_url_env, matching the Codex result.
no-key wait 90s： Fixed. If neither an env credential nor a detectable CLI login is present, the check now returns missing_api_key immediately. If a CLI login exists, it still runs a short probe with a 25s budget so login-based auth continues to work.

qurakchin (approved):

jx-qiu (approved):
Looks good. You can ask the agent to remove some not-very-useful tests like test_cli_json_output_is_parseable_and_matches_the_result. The connectivity check functionality should not require so many additional tests.

PR #160, opened 2026-09-08
feat(planner): configure context limits for Codex providers

huanming99:
## Summary

- Add optional `CODEX_MODEL_CONTEXT_WINDOW` and `CODEX_AUTO_COMPACT_TOKEN_LIMIT` overrides for the Codex planner.
- Document local-model setup for Codex and Claude Code in English and Chinese.
- Keep the vLLM `--max-model-len` example consistent with the Codex context configuration.

## Testing

- `pytest tests/unit_tests/rpent/planner/test_codex_contracts.py -q` — 25 passed
- `pre-commit run --all-files`
- English and Chinese Sphinx builds with `-W`
- Verified `rpent-check-llm` against Qwen3.6-27B served by vLLM for both Codex and Claude Code

### DIFF
--- a/docs/source-en/rst_source/usage/configure_planner.rst
+++ b/docs/source-en/rst_source/usage/configure_planner.rst
@@ -116,6 +116,27 @@ Notes:
   `Claude Agent SDK docs
   <https://code.claude.com/docs/en/agent-sdk/overview>`_.
 
+Local models with Claude Code
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Claude Code can use a local model server that implements the Anthropic
+Messages API. For a server exposing Qwen3.6-27B as
+``Qwen/Qwen3.6-27B``, configure:
+
+.. code-block:: bash
+
+   export ANTHROPIC_BASE_URL=http://127.0.0.1:8000
+   export ANTHROPIC_API_KEY=EMPTY
+
+   rpent --robot libero --planner claude_code \
+     --model Qwen/Qwen3.6-27B \
+     --suite libero_goal_task --task 1 --seed 0
+
+Claude Code assumes a 200,000-token context window for unrecognized model IDs.
+If the local server uses a different limit, see the `Claude Code environment
+variables <https://code.claude.com/docs/en/env-vars>`_ for its context and
+auto-compaction settings.
+
 .. _planner-codex:
 
 The ``codex`` planner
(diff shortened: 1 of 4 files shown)


docs/source-en/rst_source/usage/configure_planner.rst
```diff
@@ -137,6 +137,46 @@ Notes:
   ``CODEX_API_KEY``. This backend does not read ``OPENAI_BASE_URL`` or
   ``OPENAI_API_KEY``.
 
+Local models for the Codex planner
```
qurakchin: add the content of local models in Claude code

rpent/planner/codex.py
```diff
@@ -105,6 +115,8 @@ def __init__(
         self._model = model or os.environ.get("CODEX_MODEL", None)
         self._base_url = os.environ.get("CODEX_BASE_URL", None)
         self._api_key = os.environ.get("CODEX_API_KEY", None)
+        self._model_context_window = _optional_int_env(MODEL_CONTEXT_WINDOW_ENV)
```
qurakchin: use os.environ.get("xxx", None)

qurakchin: _optional_int_env can be removed

qurakchin:
can codex detect the model ctx window from vllm api?

docs/source-en/rst_source/usage/configure_planner.rst
```diff
@@ -137,6 +137,46 @@ Notes:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The Codex planner can also use a local model served through a
+Responses-compatible endpoint. For example, start a model with vLLM:
+
+.. code-block:: bash
+
+   vllm serve /path/to/model \
```
Sonorous281: The example sets CODEX_MODEL_CONTEXT_WINDOW=262144, but the vLLM command above does not set --max-model-len 262144. Could we make these consistent so the configured Codex context window always matches the limit actually exposed by vLLM?

docs/source-zh/rst_source/usage/configure_planner.rst
```diff
@@ -125,6 +125,42 @@ RPent 通过 SDK 创建进程内 MCP 服务，并把 toolkit 的工具注册到
   ``CODEX_API_KEY``；这里不读取 ``OPENAI_BASE_URL`` 或
   ``OPENAI_API_KEY``。
 
+为 Codex planner 接入本地模型
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Codex planner 也可以使用由 Responses API 兼容端点提供的本地模型。例如，可以
```
Sonorous281: The Chinese wording here feels a bit literal and mixed with English terminology.

huanming99:
I checked this with Codex 0.147.0.

vLLM returns the standard OpenAI model-list shape using `data[].max_model_len`, while the Codex model catalog expects `models[].context_window`.

With a vLLM-shaped response, Codex treated `Qwen/Qwen3.6-27B` as an unknown model and used its fallback metadata: a 272,000-token context window, with 258,400 tokens available after the 5% headroom.

Therefore Codex does not currently obtain the 262,144-token vLLM limit automatically, so the explicit context override is still needed. I updated the documentation to explain this distinction.

docs/source-en/rst_source/usage/configure_planner.rst
huanming99: Added short local-model examples for Claude Code in both the English and Chinese documentation.

No RPent code change is required for this path because Claude Code reads `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` from the child-process environment. I also verified the example against the local Qwen/vLLM service.

rpent/planner/codex.py
huanming99: Updated. The helper and additional constants were removed. The two optional values are now read directly with `os.environ.get(..., None)`.

docs/source-en/rst_source/usage/configure_planner.rst
huanming99: Updated. The vLLM example now explicitly sets `--max-model-len 262144`, matching `CODEX_MODEL_CONTEXT_WINDOW=262144`.

docs/source-zh/rst_source/usage/configure_planner.rst
huanming99: Updated. I rewrote the Chinese section with more natural wording and made the vLLM runtime limit, Codex context window, and compaction threshold easier to distinguish.

qurakchin (approved):

wilburx813 (approved):

PR #162, opened 2026-09-08
refactor: unify RLinf python path resolution and to_numpy_tree

jx-qiu:
This pull request refactors the robot environment server modules to (1) provide a unified way to start env servers using development checkouts of `rlinf`, and (2) standardize the conversion of data structures to NumPy arrays using a single `to_numpy_tree` utility.


rpent/utils/daemon.py
```diff
@@ -85,6 +94,7 @@ def __init__(
         self.cmd = cmd
         self.subprocess_env = os.environ.copy()
         self.subprocess_env.update(env_overrides or {})
+        prepend_pythonpath(self.subprocess_env, str(get_rlinf_repo_path()))
```
qurakchin: It's better to use env_overrides to wrap PYTHONPATH env

rpent/utils/serialization.py
```diff
@@ -0,0 +1,39 @@
...
+        return value.detach().cpu().numpy()
+    if dataclasses.is_dataclass(value):
+        return to_numpy_tree(dataclasses.asdict(value))
+    if isinstance(value, dict):
+        return {key: to_numpy_tree(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [to_numpy_tree(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(to_numpy_tree(item) for item in value)
+    if isinstance(value, np.generic):
```
qurakchin: why need to change np.generic to a raw value

rpent/utils/serialization.py
```diff
@@ -0,0 +1,39 @@
...
+import dataclasses
+from typing import Any
+
+import numpy as np
+
+
+def to_numpy_tree(value: Any) -> Any:
+    """Recursively convert tensors and nested values into pickle-safe data."""
+    if hasattr(value, "detach") and hasattr(value, "cpu") and hasattr(value, "numpy"):
+        return value.detach().cpu().numpy()
```
qurakchin, line 26: I think for gpu tensors, only .numpy() will copy and trans into cpu, so the .detach().cpu() is redundant.
But for cpu tensors, . detach() is not a deep copy.

jx-qiu: ``detach()`` is required because ``numpy()`` refuses tensors that
      require grad;
``cpu()`` is required because ``numpy()`` only accepts CPU tensors —
      for GPU tensors this is the call that copies to host memory;
for CPU tensors both calls are no-ops and ``numpy()`` is zero-copy:
      the returned array shares memory with the source tensor, so callers
      that later mutate the tensor in place must copy first.

rpent/utils/serialization.py
jx-qiu: Removed the cast, while making HTTP RPC preserve np generics.

rpent/utils/daemon.py
jx-qiu: Fixed.

qurakchin (approved):

Sonorous281 (approved):

PR #186, opened 2026-09-15
build(deps): allow uv prereleases for transitive explicit pins

Sonorous281:
Allow uv to resolve prerelease versions required by transitive dependencies with explicit prerelease pins.

* Change `tool.uv.prerelease` from `if-necessary` to `allow`.
* Avoid dependency resolution failures when robot/runtime extras depend on explicitly pinned prerelease packages.
* No direct dependency versions are changed.


jx-qiu (approved):
verified on my ubuntu machine.

qurakchin (approved):

PR #189, opened 2026-09-15
ci: remove duplicate documentation build

wilburx813:
Remove the duplicate documentation build workflow.


Sonorous281:
Could we also update .agents/skills/docs-check/SKILL.md to reference the English and Chinese Read the Docs configurations? It still directs agents to the deleted .github/workflows/docs.yml for CI commands.

Sonorous281 (approved):

wilburx813:
> Could we also update .agents/skills/docs-check/SKILL.md to reference the English and Chinese Read the Docs configurations? It still directs agents to the deleted .github/workflows/docs.yml for CI commands.

done

qurakchin (approved):

PR #198, opened 2026-09-17
refactor(tools): derive native tools from functions and bound methods

wilburx813:
## Summary

Tool definitions currently duplicate Python signatures in handwritten schemas and adapt results through provider-specific content blocks. This change derives native `Tool` declarations from typed functions and instance methods, and returns `ToolResult(data, images, error)` across robots, planners/MCP, and Dashboard.

- Preserve existing primitive objects, robot resource ownership, recording, and cancellation callbacks. Register bound methods with `iter_tools` / `add_tool`; use `with_handler` for resource binding and execution guards.
- Generate schemas and descriptions from annotations, `Field` constraints, and Google-style docstrings. Support `@tool`, `@tool()`, and `@tool(readonly=True)`.
- Validate arguments in Toolkit before execution or observation capture. Store accepted finish results on Toolkit and convert native results only at planner/MCP boundaries.
- Move memory path authorization into `MemoryManager` and update the English/Chinese extension guides.

## Compatibility

Custom integrations must register native declarations with `add_tool` / `add_tools`, discover them through `list_tools()`, and return `ToolResult` from handlers and observation capture. Direct Python callers use `.data` or `.to_dict()` instead of indexing the old result dictionary. Images are an ordered PNG byte list; errors use the explicit `error` field.

Historical schema snapshots cover common tools and all five robots (67 tools), retaining their published input schemas. Runtime validation rejects unknown arguments and non-finite numbers and applies declared constraints. Tool calls remain serialized. `readonly` controls automatic observation capture, including for file tools.

Adds `docstring-parser` and removes the direct `jsonschema` dependency.


rpent/tools/base.py
```diff
@@ -0,0 +1,230 @@
...
+def readonly(handler: Callable[ParamsT, ToolResult]) -> Callable[ParamsT, ToolResult]:
+    """Skip automatic environment observation capture; file writes remain allowed.
+
+    Place this marker below @tool. Calls execute one at a time.
+    """
+    setattr(handler, "_rpent_readonly", True)
+    return handler
+
+
+def tool(function: Callable[ParamsT, ToolResult], /) -> Tool[ParamsT]:
```
qurakchin: Is @tool@readonly same with @readonly@tool?

qurakchin: if not, support @tool() and @tool(readonly=True)

rpent/tools/toolkit.py
```diff
@@ -12,31 +12,37 @@
...
+from rpent.tools.base import (
+    RobotT,
+    Tool,
+    ToolContext,
+    ToolResult,
+)
+from rpent.tools.common_tools import COMMON_TOOLS
+from rpent.utils.logging import get_logger
+
+logger = get_logger("tools")
```
qurakchin, line 39: better to use the same name as the filename

rpent/tools/base.py
```diff
@@ -0,0 +1,230 @@
...
+from pydantic import BaseModel, ConfigDict, Field, create_model
+from pydantic.json_schema import GenerateJsonSchema
+from pydantic_core import core_schema
+
+if TYPE_CHECKING:
+    from rpent.memory import MemoryManager
+    from rpent.session import EnvState
+
+ParamsT = ParamSpec("ParamsT")
+RobotT = TypeVar("RobotT")
```
qurakchin: better use RuntimeT or RobotRuntimeT

robots/dual_franka/toolkit.py
```diff
@@ -391,10 +419,10 @@ def get_env_state(self, *, command, result, elapsed_s):
                 "attempt_start_step": self._attempt_start_step,
             }
             self.state.save("exploration.json", status)
-            output["exploration"] = status
+            output.data["exploration"] = status
             # Keep the original record layout, and expose lifecycle errors to the planner.
```
Sonorous281: “Keep the original record layout” will become stale once this refactor is merged. Something like Propagate lifecycle errors to the planner. is enough.

docs/source-en/rst_source/development/add_primitive.rst
```diff
@@ -32,54 +32,71 @@ From the LLM's perspective, both types expose the same interface: a
...
+values in ``data``, PNG bytes in ``images``, and failures in ``error``. A primitive
+that calls another tool directly receives the same ``ToolResult``.
+Use ``@tool(readonly=True)`` to skip automatic observation capture.
+Tools declared with ``@tool`` or ``@tool()`` capture observations by default when
+called through ``Toolkit.execute_tool``. Direct Python calls do not capture
+observations. ``readonly`` controls this capture step; it does not prohibit file
+writes or allow concurrent tool execution.
+``iter_tools`` collects decorated members from the supplied instances or modules,
+including inherited methods. Adding a primitive requires decorating its method;
+there is no separate schema or tool-name list to update. Undecorated methods and
```
Sonorous281: Could we keep this section focused on the current API instead of decorator syntax aliases and the previous schema-registration model? @tool and @tool() are equivalent for users, and “there is no separate schema or tool-name list to update” reads like migration history.

Sonorous281:
Since this PR replaces the tool declaration mechanism, the docs should focus on the current usage rather than the previous TOOLS_SPEC approach. Ideally, this replacement should not significantly increase the overall documentation length.

qurakchin (approved):

robots/dual_franka/toolkit.py
wilburx813: done

PR #203, opened 2026-09-18
feat(robots): add opt-in direct action primitive

Elessar123:
Agents can use VLA and scripted primitives such as `move_to`, but cannot submit a native environment action as a planner tool. This adds `execute_action(values=[...])` alongside those primitives for LIBERO, RoboCasa, and RoboTwin, exposed only when the run starts with `--enable-direct-action`.

### Environment-owned action contract

- The connected environment supplies native action layouts and per-coordinate bounds through the read-only `env.get_action_spec` RPC. The shared client generates the tool schema and validates inputs from that specification; the primitive has no fixed action dimensions or blanket `[-1, 1]` limit.
- LIBERO and RoboCasa read the active simulator's native `action_spec`. RoboTwin describes native joint/pose waypoint layouts from its configured robot. Typed environments advertise their supported `action_type` values and default mode.
- Existing environment clients, cancellation, observation updates, recordings, and recipe export are reused. RoboCasa invalidates VLA history after a direct action so subsequent VLA calls reseed from the current scene.
- Default runs do not register the new tool or query its specification. No new dependencies. External environment servers must be updated with the client before enabling this optional tool.
- Paired README documentation and offline tests cover enablement, action forwarding, different environment dimensions and bounds, invalid inputs, cancellation, state records, and VLA history continuity.

This makes the direct-action primitive independent of a particular robot's vector size. It does not broaden the set of robot/controller configurations supported by the existing execution backends.

### Validation

- `pre-commit run --all-files` and `pre-commit run --files rpent/robots/components/action_spec.py` — passed.
- `.venv/bin/python -m pytest tests/unit_tests/robots/test_direct_action.py -q --tb=short` — **36 passed**.
- `.venv/bin/python -m pytest tests/unit_tests -q --tb=short` — **613 passed, 3 skipped, 4 failed**. All four failures are in `test_pi05_vla_server_contracts.py` because the test environment lacks `torch`; the same four failures reproduced on unmodified main.
- Manual simulation smoke tests on the revised implementation used the changed checkout, production robot runtime hooks with only the environment component, real RPC clients, and `Toolkit.execute_tool("execute_action", ...)` on GPU 3. Each test fetched the live environment action specification:
  - **LIBERO-PRO**, `libero_object_swap`, task 2, seed 0: five actions moved the end effector **16.42 mm**; six state records, updated images, and an episode video were produced.
  - **RoboCasa**, `OpenDrawer`, target split, seed 0: eight actions moved the base **1.59 mm**; nine state records, updated images, and an episode video were produced.
  - Both owned environment processes shut down successfully.

The simulation checks establish action execution and observation/artifact updates, not autonomous planner performance or benchmark task success. RoboTwin simulation execution and live VLA/direct-action alternation remain unverified. RoboTwin native-layout tests require its optional runtime and were skipped locally; RoboCasa's VLA-history reseeding is covered offline.


README.md
```diff
@@ -173,6 +173,20 @@ rpent --robot libero --suite libero_object_swap --task 2 --seed 0 \
...
+expose the optional `execute_action(values=[...])` primitive alongside VLA and
+scripted tools. It executes one native environment action per call and records
+the resulting state. Without the flag, the tool is not registered.
+
+The tool reads action dimensions, per-coordinate bounds, and supported action
+modes from the connected environment. Supply values in that environment's native
+controller order; the tool description includes its current specification.
+RoboTwin advertises its native joint and end-effector waypoint modes.
+When attaching an external environment server, update it together with the client:
+this optional tool requires the `env.get_action_spec` RPC.
```
QuanluZhang, line 188: this part should not be placed in README, it should be placed in doc, for example, in planner.

robots/libero/robot_spec.py
```diff
@@ -173,6 +174,11 @@ def _add_cli_args(parser: argparse.ArgumentParser, use_dashboard: bool) -> None:
     usual usage message.
     """
     required = not use_dashboard
+    parser.add_argument(
```
jx-qiu: Should we enable direct action or not, by default? In which cases do we want to turn off the direct action functionality?

